Issue #33, opened 2018-05-25
Feature/management endpoint to list subjects


Jakedavies:
<a href="https://github.com/sindelio"><img src="https://avatars1.githubusercontent.com/u/19046387?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [sindelio](https://github.com/sindelio)**
_Wednesday May 23, 2018 at 16:19 GMT_
_Originally opened as https://github.com/clevertech/OpenGDPR/pull/236_

----

**Remade the branch for a cleaner history**
Add management endpoint to list users and its tests.
Working nicely, closes #189 

----

_**[sindelio](https://github.com/sindelio)** included the following code: https://github.com/clevertech/OpenGDPR/pull/236/commits_

Jakedavies:
<a href="https://github.com/Jakedavies"><img src="https://avatars1.githubusercontent.com/u/3431461?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [Jakedavies](https://github.com/Jakedavies)**
_Wednesday May 23, 2018 at 22:12 GMT_

----

@sindelio please do not close the old PR and open a new one, it makes it harder to re-review when we ask for changes since we do not know what changes we requested were anymore

Jakedavies:
<a href="https://github.com/sindelio"><img src="https://avatars1.githubusercontent.com/u/19046387?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [sindelio](https://github.com/sindelio)**
_Thursday May 24, 2018 at 02:32 GMT_

----

Sorry for that, I tried rebasing the older branch on top of feature/open-source-prep-rebased but it ended with duplicated commits. Also there were a lot of coflicts to cherry pick. I probably did something funny in that rebase, that's why i remade the branch. Will try to not recreate PR's from now on

PR #52, opened 2018-05-29
Management endpoint to list subjects and its tests, plus some cleanup

### LINKED ISSUE #18: Create a management endpoint for listing users in the system with their data
<a href="https://github.com/Jakedavies"></img></a> **Issue by [Jakedavies](https://github.com/Jakedavies)**
_Monday May 14, 2018 at 23:13 GMT_
_Originally opened as https://github.com/clevertech/OpenGDPR/issues/189_

----

This will feed into the admin dashboard listing of users. 

There was talk about what data we are allowed to expose and what data we can't, for now lets format all data fields out of the system like  `"{name: h****h, email: "h***@*****.com" }"` or similar.

sindelio:
Tests working isolated and in batch.
Closes #18 #33

### DIFF
--- a/cg/test/management/subjects.api.spec.js
+++ b/cg/test/management/subjects.api.spec.js
@@ -1,0 +1,231 @@
+const { initResources, fetch, closeResources } = require('../utils');
+const { db } = require('../../src/db');
+const { generateClientKey, encryptForStorage, hash } = require('../../src/utils/encryption');
+const { managementJWT } = require('../../src/utils/jwt');
+const { Unauthorized } = require('../../src/utils/errors');
+
+beforeAll(initResources);
+afterAll(closeResources);
+
+describe('List subjects that have given consent', () => {
+  it('should not allow a manager to list subjects without a manager JWT', async () => {
+    //GIVEN AND WHEN
+    const res = await fetch('/api/management/subjects/list', {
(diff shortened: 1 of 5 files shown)


cg/test/management/subjects.api.spec.js
```diff
@@ -0,0 +1,231 @@
...
+      headers: {
+        Authorization: `Bearer ${managementToken}`
+      }
+    });
+
+    //THEN
+    expect(res.ok).toBeTruthy();
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual(
+      expect.arrayContaining([])
```
maksymilian-majer: I think this will be non-deterministic, since other tests can fill some other subjects. We should expect, that the array does not contain the particular subject, that has been added in this test.

cg/test/management/subjects.api.spec.js
```diff
@@ -0,0 +1,231 @@
...
```
maksymilian-majer: Same as above.

cg/test/management/subjects.api.spec.js
```diff
@@ -0,0 +1,231 @@
...
```
maksymilian-majer: and here

PR #54, opened 2018-05-30
api is now hashing users with bcrypt

Jakedavies:


api/test/users.spec.js
```diff
@@ -6,6 +6,28 @@ beforeAll(initResources);
 afterAll(closeRources);
```
sindelio: There's a typo here xD

PR #69, opened 2018-06-01
Feature/interactive setup script

### LINKED ISSUE #42: Interactive quick start script
The Quick Start guide can be mostly converted into a wizard like interactive shell script and generate the right config, .env files, docker-compose settings, etc.

Relates to #24

### LINKED ISSUE #71: Create a named volume for Postgres
Right now we're relying on a default volume created for the postgres docker images and when we want to destroy the volumes there's no convenient method (ex. `docker-compose down -v`).
We should add a named volume for postgres data, just like we do with quorum containers.

Jakedavies:
closes #42, closes #71

### DIFF
--- a/setup.js
+++ b/setup.js
@@ -1,0 +1,55 @@
+const crypto = require('crypto');
+var exec = require('child_process').exec;
+
+const readline = require('readline');
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout
+});
+
+function getKey() {
+  return new Promise((resolve, reject) => {
+    crypto.randomBytes(48, function(err, buffer) {
+      if (err) reject(err);
+      var token = buffer.toString('hex');
+      resolve(token);
+    });
+  });
+}
+
+function getInput(question) {
+  return new Promise(resolve => {
+    rl.question(question, answer => {
+      resolve(answer);
+    });
+  });
+}
+
+function checkPrereqs() {
+  // check docker
+}
+
+// const args = process.argv.slice(2);
+
+async function run() {
+  await checkPrereqs();
+  const defaultAccountPassword = await getKey();
+  const ACCOUNT_PASSWORD = getInput(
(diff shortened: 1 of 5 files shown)


Jakedavies:
@maksymilian-majer would be good if you could run this and see if everything works for you okay

setup.js
```diff
@@ -0,0 +1,114 @@
...
+  )}`.replace(/\n$/, '');
+  const processorAccount = `0x${exec('cat quorum/generated_configs/node2/dd/account.txt')}`.replace(
+    /\n$/,
+    ''
+  );
+
+  let dbPassword = await getInput(`Enter a db password: default [${defaultDbPassword}]`);
+  if (!dbPassword) dbPassword = defaultDbPassword;
+
+  logProgress('Generating config for CG controller and CG processor?');
```
maksymilian-majer: I left a `?` there accidentally

setup.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/* eslint max-statements: 0 */
+async function run() {
+  await checkPrereqs();
+  const defaultAccountPassword = await getKey();
+  const defaultDbPassword = await getKey();
+
+  let accountPassword = await getInput(
+    `Enter an account password: default [${defaultAccountPassword}]`
```
maksymilian-majer: The user's input blends in with the message. Add a space at the end afte `]`?

setup.js
```diff
@@ -0,0 +1,114 @@
...
+
+  logProgress('Deploying initial contract...');
+
+  exec(`docker/compose exec -T cg yarn run deploy-contract`);
+
+  logProgress('Adding initial processor');
+
+  exec(`docker/compose exec -T cg yarn run add-processor ${processorAccount}`);
+
+  logProgress('All done, the example UI should be accessible at http://localhost:3000');
```
maksymilian-majer: For some reason the script doesn't exit after that. An open stream?

PR #83, opened 2018-06-04
creating a temp db for each test

### LINKED ISSUE #68: Create test DB per test file instead global
Currently we have a global DB setup/teardown, but it's causing conflicts between test suits/
We should experiment in having DB per test file and see if that does not affect test performance too much (although in test feedback building the docker-compose setup is the longest part anyway, so if we optimise that, we can add a few seconds to the tests).

maksymilian-majer:
Closes #68


cg/knex-config-factory.js
```diff
@@ -0,0 +1,22 @@
+module.exports = function() {
+  const dbName = global.dbName || process.env.DB_DATABASE;
```
maksymilian-majer: this is a hack, so that we can inject dbName in tests – I'm open to better solutions

cg/test/random-test-db-environment.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = class RandomDbEnvironment extends NodeEnvironment {
+  constructor(config) {
+    super(config);
+  }
+
+  async setup() {
+    await super.setup();
+
+    this.global.dbName = `${process.env.DB_DATABASE}_${getRandomId()}`;
```
maksymilian-majer: This global is only set for tests, otherwise it will be read from ENV vars (hack continued)

PR #86, opened 2018-06-04
Fixes the default startup command from dockerfile

### LINKED ISSUE #87: Wrong startup command in frontend dockerfile
Kubernetes will by default use the CMD from the dockerfile if it is not overridden, the one in our dockefile is wrong

Jakedavies:
Closes #87 

Kubernetes appears to be using the default startup command from the dockerfile(which doesn't work), this changes it to use the one we normally use from docker-compose.yml

This is starting the server with webpack, which isn't a very production ready method but this is just the demo UI so its not a big deal.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -7,10 +7,10 @@ RUN apk --no-cache add ca-certificates && update-ca-certificates
 # Force python update to install Sass
 RUN apk upgrade && \
   apk add --no-cache git \
-    make gcc g++ python
+  make gcc g++ python
 
 EXPOSE 3000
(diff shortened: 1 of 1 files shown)


Jakedavies:
While this will fix the issue, there is a better solution already built, we must have screwed up the code that does our final build. We should be using the dockerfile.prod for frontend production builds rather than the development one.

We probably don't want to merge this, but want @maksymilian-majer to review Martino's suggested solution first

maksymilian-majer:
I think #89 will solve this, so we should hold off on merging.

maksymilian-majer:
I was wrong – it's good to merge @Jakedavies

PR #88, opened 2018-06-04
Feature/basic stats for management panel

Jakedavies:
I designed this around #68 being ready, since it hard to test otherwise, do not merge until that is merged into master and we rebase this on top

### DIFF
--- a/cg/src/domains/management/stats/stats.service.js
+++ b/cg/src/domains/management/stats/stats.service.js
@@ -28,10 +28,18 @@ class StatsService {
       .select(db.raw('count(subjects.id) as subject_count'))
       .select('subject_processors.processor_id');
 
+    const processors = await db('processors');
+
     const processorData = processorsWithSubjectCount.reduce((current, processor) => {
       current[processor.processor_id] = { consented: parseInt(processor.subject_count, 10) };
       return current;
     }, {});
+
+    processors.forEach(processor => {
+      if (!processorData[processor.id]) {
+        processorData[processor.id] = { consented: 0 };
+      }
+    });
 
     return {
       controller: controllerData,
--- a/cg/src/domains/management/stats/stats.service.js
+++ b/cg/src/domains/management/stats/stats.service.js
@@ -7,12 +7,12 @@ class StatsService {
 
   async stats() {
     const [result] = await db('subjects')
-      .select(db.raw('count subjects.id as consented'))
+      .select(db.raw('count(subjects.id) as consented'))
       .join('subject_keys', 'subjects.id', 'subject_keys.subject_id');
-    const consentedSubjectCount = result.consented;
+    const consentedSubjectCount = parseInt(result.consented, 10);
 
-    const [result2] = await db('subjects').select(db.raw('count subjects.id as total'));
-    const totalSubjectCount = result2.total;
+    const [result2] = await db('subjects').select(db.raw('count(subjects.id) as total'));
+    const totalSubjectCount = parseInt(result2.total, 10);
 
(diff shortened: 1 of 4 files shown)


cg/test/management/stats.api.spec.js
```diff
@@ -0,0 +1,107 @@
+jest.mock('../../src/utils/blockchain/web3-provider-factory');
+
+const winston = require('winston');
+
+const { initResources, fetch, closeResources } = require('../utils');
+const { deployContract } = require('../blockchain-setup');
+const { db } = require('../../src/db');
+const { managementJWT, subjectJWT } = require('../../src/utils/jwt');
+const { setProcessors } = require('../../src/utils/blockchain');
```
sindelio: Just a reminder that setProcessors is meant to facilitate tests and should not be used in the actual API

cg/test/management/stats.api.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        data: {
+          controller: {
+            consented: 1,
+            unconsented: 1,
+            total: 2
+          },
+          processors: {
+            '201': {
+              // I was expecting this to be 1 originally, but it appears the be 2 because requesting erasure does not cascade, should it?
+              consented: 2
```
sindelio: A processor can't have consent for a subject's data that was erased, that's misleading. I think we should address this in the erase-data endpoint

sindelio: Also, I can't think of other tests for this endpoint

Jakedavies:
@sindelio I will create an issue to correct the behavior of the erase-data functionality since it is probably out of scope for this PR

cg/src/domains/management/stats/stats.service.js
```diff
@@ -0,0 +1,43 @@
...
+
+class StatsService {
+  constructor(database = db) {
+    this.db = database;
+  }
+
+  async stats() {
+    const [result] = await db('subjects')
+      .select(db.raw('count(subjects.id) as consented'))
+      .join('subject_keys', 'subjects.id', 'subject_keys.subject_id');
```
maksymilian-majer: We could have one query instead of two to left join on keys and count when not null (use `COALESCE` with `CASE` or something like it), but I don't expect a huge performance gain.

cg/test/management/stats.api.spec.js
```diff
@@ -0,0 +1,107 @@
...
+  addressesToInsert.forEach(a => expect(addresses).toContainEqual(a));
+});
+
+beforeEach(() => {
+  process.env.MODE = VALID_RUN_MODES.CONTROLLER;
+});
+afterAll(closeResources);
+
+describe('Stats endpoint', () => {
+  it('Should display the stats for the current system state', async () => {
```
maksymilian-majer: I suggest a test case (at least one) with no data first to see if we get zero's. I suspect a 500 error when there are no subjects. Maybe also some edge cases when consented is 0 and total is > 0, etc.

cg/src/domains/management/stats/stats.service.js
```diff
@@ -0,0 +1,43 @@
...
+    const consentedSubjectCount = parseInt(result.consented, 10);
+
+    const [result2] = await db('subjects').select(db.raw('count(subjects.id) as total'));
+    const totalSubjectCount = parseInt(result2.total, 10);
```
maksymilian-majer: If there are no subjects I suspect there will be an error thrown on `result2.total`.

maksymilian-majer: Never mind – count will always return result even if zero.
Now there's a test for this :)

PR #97, opened 2018-06-06
Avoid to pass query string to the callback url

### LINKED ISSUE #58: After signing up you should not be redirected with all the POST params the our url
After signing up for the clevertech times i was redirected to `http://localhost:3000/success?firstname=test&lastname=test&website=test&email=d%40g.com&password=password&repeatpassword=password`

This is obviously not desirable for a number of reasons, lets fix this

fjcero:
When redirecting in the `give consent` flow, the current callback URL is `/success` (this could be improved though), and right now the query strings are cleaned up for either GET or POST method.

Closes #58

### DIFF
--- a/frontend/src/containers/SignUp.js
+++ b/frontend/src/containers/SignUp.js
@@ -13,6 +13,11 @@ class SignUp extends Component {
   onSubmit(e) {
     e.preventDefault();
     this.setState({ isLoading: true });
+    setTimeout(() => {
+      window.location = '/success';
+      // this.props.history.push(`/success`);
+    }, 1500);
+    return false;
   }
 
   render() {
@@ -27,6 +32,7 @@ class SignUp extends Component {
           fontSize: '12px'
         }
       },
+      callbackUrl: '/success',
       required: true
     });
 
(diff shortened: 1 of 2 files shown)


frontend/src/containers/SignUp.js
```diff
@@ -13,6 +13,11 @@ class SignUp extends Component {
   onSubmit(e) {
     e.preventDefault();
     this.setState({ isLoading: true });
+    setTimeout(() => {
```
Jakedavies: this is a pretty weird solution, can you explain to me what you think was going wrong and how this fixes it?

fjcero: You are right. `setTimeout` is not necessary, just wanted to show the `loading` effect of the button because backend in localhost is too fast. The redirect is part of solving the issue. If we provide a custom element for a form, we should provide some way to programmatically handle the form success, with a callback or something. Since this is just a demo, I forced the "callback" to the "ConsentSuccess" component. In the future, we should provide something like this, the same way Auth0 does.

frontend/src/elements/components/Consent.js
```diff
@@ -72,6 +72,10 @@ class Consent extends React.PureComponent {
     let data = {};
     let processors = this.state.processors.slice();
 
+    if (processors.length < 1 && !ReactDOM.findDOMNode(this)) {
```
maksymilian-majer: Why is that?
We should be able to consent to only the controller.

fjcero: Hm, that's an interesting question then, I will modify to allow that, makes sense.

frontend/src/containers/SignUp.js
Jakedavies: Lets chat about this in the meeting tomorrow I want to make sure we aren't band aiding a deeper issue here by forcing a redirect with window.location

PR #103, opened 2018-06-06
Coverup/Blanking for sensitive subjects data displayed in the management dashboard

sindelio:
Tested in Postman and via our test suite. 
There was no need to make a specific test for the covering up of the data because it is already part of the expectation in other tests. Although we could make one for clarity, what do you guys think?

### DIFF
--- a/cg/src/domains/management/subjects/subjects.controller.js
+++ b/cg/src/domains/management/subjects/subjects.controller.js
@@ -4,11 +4,11 @@ const _ = require('underscore');
 function coverString(string) {
   const firstCharacter = string[0];
   const lastCharacter = string[string.length - 1];
-  let coveredString = new Array(string.length).fill('*');
-  coveredString[0] = firstCharacter;
-  coveredString[string.length - 1] = lastCharacter;
-  coveredString = coveredString.join('');
-  return coveredString;
+  const fullyCoveredArray = new Array(string.length).fill('*');
+  fullyCoveredArray[0] = firstCharacter;
+  fullyCoveredArray[string.length - 1] = lastCharacter;
+  let partiallyCoveredString = fullyCoveredArray.join('');
+  return partiallyCoveredString;
 }
 
 class SubjectsController {
--- a/cg/src/domains/management/subjects/subjects.controller.js
+++ b/cg/src/domains/management/subjects/subjects.controller.js
@@ -7,7 +7,7 @@ function coverString(string) {
   const fullyCoveredArray = new Array(string.length).fill('*');
   fullyCoveredArray[0] = firstCharacter;
   fullyCoveredArray[string.length - 1] = lastCharacter;
-  let partiallyCoveredString = fullyCoveredArray.join('');
+  const partiallyCoveredString = fullyCoveredArray.join('');
   return partiallyCoveredString;
 }
 
--- a/cg/src/domains/management/subjects/subjects.controller.js
+++ b/cg/src/domains/management/subjects/subjects.controller.js
@@ -1,4 +1,15 @@
 const SubjectsService = require('./subjects.service');
+const _ = require('underscore');
+
+function coverString(string) {
+  var firstCharacter = string[0];
+  var lastCharacter = string[string.length - 1];
+  var coveredString = new Array(string.length).fill('*');
+  coveredString[0] = firstCharacter;
+  coveredString[string.length - 1] = lastCharacter;
+  coveredString = coveredString.join('');
+  return coveredString;
+}
 
 class SubjectsController {
   constructor(subjectsService = null) {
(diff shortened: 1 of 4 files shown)


cg/src/domains/management/subjects/subjects.controller.js
```diff
@@ -7,7 +18,16 @@ class SubjectsController {
 
   async listSubjects(req, res) {
     const subjectsPage = await this.subjectsService.listSubjects(req.query.page);
-    res.json(subjectsPage);
+    if (subjectsPage.hasOwnProperty('error')) {
```
sindelio: I need to fix this line, although it's working, a subject can insert an 'error' key in his personal data and that would make his data not be blanked when displayed in the management dashboard. I can think of 2 fixes: 
- Reserve the error key to the system, not allowing a subject's personal data to have an  error key
- Do a more strict check, with the contents of the error field. this does not really solve the issue, but will make it less likely to the point that only someone trying to hack the system will be able to do it
I want to hear your opinions on this

cg/src/domains/management/subjects/subjects.controller.js
```diff
@@ -1,4 +1,15 @@
 const SubjectsService = require('./subjects.service');
+const _ = require('underscore');
+
+function coverString(string) {
+  var firstCharacter = string[0];
```
Jakedavies: use let

Jakedavies: actually all these variables should be const

cg/src/domains/management/subjects/subjects.controller.js
```diff
@@ -1,4 +1,15 @@
...
+  var lastCharacter = string[string.length - 1];
+  var coveredString = new Array(string.length).fill('*');
+  coveredString[0] = firstCharacter;
+  coveredString[string.length - 1] = lastCharacter;
+  coveredString = coveredString.join('');
```
Jakedavies: use a  new variable here, no reason to overwrite

cg/src/domains/management/subjects/subjects.controller.js
Jakedavies: Why are we checking the output of the service layer for an error object at all? I would expect that any error case at the service layer is going to throw an error and then bubble out to the top level?

sindelio: This shows that I don't understand well how we handle errors in the project yet. I will try that out

cg/src/domains/management/subjects/subjects.controller.js
```diff
@@ -1,4 +1,15 @@
 const SubjectsService = require('./subjects.service');
+const _ = require('underscore');
+
+function coverString(string) {
+  const firstCharacter = string[0];
+  const lastCharacter = string[string.length - 1];
+  let coveredString = new Array(string.length).fill('*');
```
Jakedavies: This is mostly a small note but I would like you to fix it so you understand. 

You can make this `coveredString` variable a const, array items can be reassigned with a const

cg/src/domains/management/subjects/subjects.controller.js
```diff
@@ -1,4 +1,15 @@
...
+  coveredString[0] = firstCharacter;
+  coveredString[string.length - 1] = lastCharacter;
+  coveredString = coveredString.join('');
```
Jakedavies: coveredString actually becomes a completed different object here (it goes from an array to a string), so we should not re-use the variable 

In the interest of code clarity and readability I would suggest you create a new variable at the stage you are getting the final string, and return that.

Jakedavies: Using a new variable here also allows the variable above you declared with `let` to be a `const`

sindelio: Hmm I see, it's interesting that I can change a const array after the first assignment, didn't knew that. I really want to comprehend more how Javascript works

cg/src/domains/management/subjects/subjects.controller.js
```diff
@@ -1,4 +1,15 @@
 const SubjectsService = require('./subjects.service');
+const _ = require('underscore');
+
+function coverString(string) {
+  const firstCharacter = string[0];
+  const lastCharacter = string[string.length - 1];
+  const fullyCoveredArray = new Array(string.length).fill('*');
+  fullyCoveredArray[0] = firstCharacter;
+  fullyCoveredArray[string.length - 1] = lastCharacter;
+  let partiallyCoveredString = fullyCoveredArray.join('');
```
Jakedavies: this should be const since it never gets re-assigned

Jakedavies:
@sindelio go ahead and merge after you change that let to a const

sindelio:
Ok

PR #104, opened 2018-06-06
added a script that auto deploys the docs on release, removed .docs

Jakedavies:
The only important changes are the .travis.yml and the deploy-docs.sh files


Jakedavies:
I have tested this locally and am _pretty_ sure it will work on travis, but will have to confirm once you approve this maks, testing on travis is a pain

Jakedavies:
Currently blocked on FTP issues with travis, trying to get dev-ops to move docs somewhere where we can use SFTP for deployment because ftp is :poop:

maksymilian-majer:
@Jakedavies you'll need to resolve conflicts and we'll need to add a test tag to check if deployment works before we merge

PR #116, opened 2018-06-09
Feature/data shares

### LINKED ISSUE #79: Create API endpoint for creating a data sharing url
This should create a unique token and store it with a name

Jakedavies:
closes #79

### DIFF
--- a/cg/src/domains/subjects/data-shares.controller.js
+++ b/cg/src/domains/subjects/data-shares.controller.js
@@ -1,0 +1,23 @@
+const DataShareService = require('./data-share.service');
+
+class DataShareController {
+  constructor(processorsService = null) {
+    this.service = processorsService || new DataShareService();
+  }
+
+  async getDataShares(req, res) {
+    res.json(await this.service.getDataShares(req.subject.id));
+  }
+
+  async createDataShare(req, res) {
+    await this.service.createDataShare(req.body.name);
+    res.json({ success: true });
+  }
+
+  async removeDataShare(req, res) {
+    await this.service.removeDataShare(req.params.dataShareId);
+    return res.json({ success: true });
+  }
+}
+
+module.exports = DataShareController;
--- a/cg/src/domains/subjects/data-shares.controller.js
+++ b/cg/src/domains/subjects/data-shares.controller.js
@@ -1,4 +1,4 @@
-const DataShareService = require('./data-share.service');
+const DataShareService = require('./data-shares.service');
 
 class DataShareController {
   constructor(processorsService = null) {
@@ -10,7 +10,7 @@ class DataShareController {
   }
 
   async createDataShare(req, res) {
-    await this.service.createDataShare(req.body.name);
+    await this.service.createDataShare(req.subject.id, req.body.name);
     res.json({ success: true });
   }
 
--- a/cg/src/domains/subjects/data-shares.controller.js
+++ b/cg/src/domains/subjects/data-shares.controller.js
@@ -1,8 +1,8 @@
 const DataShareService = require('./data-shares.service');
 
 class DataShareController {
-  constructor(processorsService = null) {
-    this.service = processorsService || new DataShareService();
+  constructor(dataShareService = null) {
+    this.service = dataShareService || new DataShareService();
   }
 
   async getDataShares(req, res) {
--- a/cg/src/domains/subjects/data-shares.service.js
+++ b/cg/src/domains/subjects/data-shares.service.js
@@ -1,0 +1,26 @@
+const { db } = require('../../db');
+const crypto = require('crypto');
+
+class DataShareService {
+  async getDataShares(subjectId) {
+    return db('data_shares').where({ subject_id: subjectId });
+  }
+
+  async createDataShare(subjectId, name) {
+    const token = crypto.randomBytes(48).toString('hex');
+    await db('data_shares').insert({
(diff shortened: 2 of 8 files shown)


cg/src/domains/subjects/data-shares.controller.js
```diff
@@ -0,0 +1,23 @@
+const DataShareService = require('./data-shares.service');
+
+class DataShareController {
+  constructor(processorsService = null) {
```
sindelio: shouldn't this be a dataShareService instead of prcessorsService?

cg/src/utils/errors.js
```diff
@@ -42,6 +42,7 @@ class NotFound extends Error {
 }
 
 const apiErrorHandler = (err, req, res, next) => {
+  console.log(err);
```
sindelio: The intention is leaving this console.log here?

cg/test/subjects/data-shares.api.spec.js
```diff
@@ -0,0 +1,140 @@
...
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${token}`
+        }
+      },
+      overrides
+    )
+  );
+}
+beforeAll(async () => {
```
sindelio: Is it ok to have 2 beforeAll() and initResources()?

sindelio:
What does it mean to create a data share in the system?

cg/test/subjects/data-shares.api.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    method: 'POST',
+    body: { personalData: { name: 'test' } },
+    headers: {
+      Authorization: `Bearer ${token}`
+    }
+  });
+  return token;
+}
+
+function fetchWithAuthorization(path, token, overrides) {
```
sindelio: What is the difference between this fetch and the other we were using?

cg/src/domains/subjects/data-shares.service.js
```diff
@@ -0,0 +1,31 @@
...
+    const data = {
+      subject_id: subjectId,
+      token,
+      name
+    };
+    await db('data_shares').insert(data);
+  }
+  async removeDataShare(dataShareId) {
+    const [dataShare] = await db('data_shares').where({ id: dataShareId });
+    if (!dataShare) throw new BadRequest('Data Share not found');
```
maksymilian-majer: Shouldn't we rather throw `NotFound()`??

cg/src/domains/subjects/data-shares.service.js
```diff
@@ -0,0 +1,31 @@
...
+    await db('data_shares')
+      .delete()
```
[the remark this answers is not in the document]
maksymilian-majer: The order does not matter and it could even be more SQL-like:
```
db.delete().from('data_shares').where({ id: dataShareId });
```
which reads like SQL:
```
DELETE FROM data_shares WHERE id = dataShareId
```

cg/test/subjects/data-shares.api.spec.js
maksymilian-majer: It is a wrapper that adds the `Authorization` header. @Jakedavies maybe we should extract it to `utils`, cause it's used all over the place?

cg/test/subjects/data-shares.api.spec.js
maksymilian-majer: I think this is a mistake

cg/test/subjects/data-shares.api.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    await deployContract();
+  } catch (e) {
+    console.error(`Failed deploying contract ${e.toString()}`);
+  }
+  await initResources();
+});
+
+describe('Listing data-shares', () => {
+  it('should return the data-shares', async () => {
+    const subjectId = '673211';
```
maksymilian-majer: since we have DB per test file we don't need to invent these random subjectIds anymore :)
that's cosmetics – I don't mean we should change it here

cg/test/subjects/data-shares.api.spec.js
```diff
@@ -0,0 +1,140 @@
...
+beforeAll(async () => {
+  try {
+    await deployContract();
+  } catch (e) {
+    console.error(`Failed deploying contract ${e.toString()}`);
+  }
+  await initResources();
+});
+
+describe('Listing data-shares', () => {
```
maksymilian-majer: I'd add a test case for listing data shares that returns an empty array – jut to be sure there's no HTTP 500

cg/test/subjects/data-shares.api.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+  it('Should error if the share does not exist', async () => {
+    const subjectId = '8422291211';
+    const token = await createUser(subjectId);
+
+    const res = await fetchWithAuthorization(`/api/subject/data-shares/439101111/remove`, token, {
+      method: 'POST'
+    });
+
+    expect(res.status).toEqual(BadRequest.StatusCode);
```
maksymilian-majer: As above – `BadRequest` vs `NotFound`?

cg/test/subjects/data-shares.api.spec.js
Jakedavies: yeah I think abstracting it makes sense, I wanted to play with it here to see if It saved any time.

PR #123, opened 2018-06-11
Admin Log In

### LINKED ISSUE #75: Hook up management UI login

fjcero:
Closes #75 

- [x] Login Form
- [x] No Bulma required


admin/public/index.html
```diff
@@ -20,6 +20,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>ClearGDPR - Management</title>
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bulma/0.7.1/css/bulma.min.css">
```
maksymilian-majer: Did @juanmnl also use bulma? Do we want it?

admin/src/components/Login/SignUp.js
```diff
@@ -0,0 +1,47 @@
...
+
+export default class SignUp extends React.PureComponent {
+  state = {
+    isLoading: false,
+    name: null,
+    email: null,
+    password: null
+  };
+
+  onSignUp(e) {
```
maksymilian-majer: Shouldn't that be split up into a container?

The SignUp component should only contain the UI code IMO and there should be a container that passes the state down and handles `onSubmit`.

admin/public/index.html
fjcero: I used Bulma since it's the boilerplate used in CT, but added it to avoid focusing on common components styling. Probably, the best way is to use it with Sass and remove not used dependencies like rebooting or grids.

Jakedavies: I think we can remove this, juans style is standalone.

admin/src/components/Login/SignUp.js
Jakedavies: I think I disagree with separating into containers/components pre-maturely, its mostly adds indirection with no benefit IMO.

Jakedavies:
I had a bit of an oversight here with the tasks.. Management UI does not need any form of sign up, admins will create other admin users, so we only need the login.

admin/src/components/Login/SignUp.js
Jakedavies: I have thought about it a bit more, you are probably right creating the "reacty" container/component separation I think makes sense especially as the complexity of the components grows.

fjcero:
I will move SignUp to do Login. I'm ok with removing Bulma, and addressing later improvements on styling, makes sense to respect @juanmnl as styleguide boilerplate.

admin/src/components/Login/Login.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+export default class SignUp extends React.PureComponent {
+  state = {
+    isLoading: false,
+    name: null,
+    email: null,
+    password: null
+  };
+
+  onSignUp(e) {
```
Jakedavies: this component is called login but calls register and refers to everything as sign up?

admin/src/config.js
```diff
@@ -0,0 +1,15 @@
+const envs = {
+  development: {
+    API_URL: 'http://localhost:8082'
```
Jakedavies: we can't really config the admin API like this, it needs to switch based on current URL. See how the frontend determines the backend url

admin/src/containers/Login/Auth.js
```diff
@@ -0,0 +1,45 @@
+import cookie from 'cookie';
+import config from '../../config';
```
Jakedavies: is this a container? It doesn't seem to render anything?

admin/src/components/Login/Login.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+export default class Login extends React.Component {
```
Jakedavies: does our linter not enforce specifying proptypes? I feel we should set it too if it is not already.

fjcero: It's not specifying right now, I can add this configuration, perhaps I missed that in the `admin` and is working in `frontend`, but not remember to be enforced to that so far. I will add it, either way, I like that.

admin/src/components/Login/Login.test.js
fjcero: Since I created the test in another branch, with the setup of jest also, will remove this for now.

admin/src/containers/Login/Auth.js
fjcero: It's more like a helper function. Is a dependency for Login container but could be used in other containers to get logged user information. Could be move to a `lib` directory, perhaps is more accurate.

fjcero:
@maksymilian-majer yup will keep working on test branch today

admin/src/components/Login/Login.js
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+  onLogin(e, auth) {
+    e.preventDefault();
+    this.setState({ isLoading: true });
+    auth(this.refs.email.value, this.refs.password.value);
+  }
+
+  render() {
+    const { auth } = this.props;
```
Jakedavies: this should be named onLoginSubmit or handleLoginSubmit, I want to enforce a bit of standardization of function naming through our react code at least

admin/src/containers/Login/Login.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import logo from '../../assets/logo.png';
+import Login from '../../components/Login/Login';
+import Session from '../../helpers/Session';
+import config from '../../config';
+import '../../theme/Login.css';
+
+const session = new Session();
+
+const login = (username, password) => {
```
Jakedavies: handleLogin

admin/src/containers/Login/Login.js
```diff
@@ -0,0 +1,38 @@
...
+    .catch(err => {
+      return err;
+    });
+};
+
+const LoginContainer = () => (
+  <section className="login-section">
+    <div className="login-card">
+      <img className="logo" src={logo} alt="Clear logo" />
+      <Login auth={login} />
```
Jakedavies: `onLoginSubmit={handleLogin}`

admin/src/components/Login/Login.js
Jakedavies: `onLoginSubmit`

PR #132, opened 2018-06-13
Add endpoint to list managers and tests for it

sindelio:
Postman and tests working, good for review

### DIFF
--- a/cg/src/domains/management/users/users.helpers.js
+++ b/cg/src/domains/management/users/users.helpers.js
@@ -8,9 +8,9 @@ const initialPassword = 'clearGDPR';
 async function ensureInitialManagementUser() {
   // not sure if the initial user creds should be hardcoded or env variables... both feel bad
   const user = await usersService.getUserByUsername('admin');
-  if (user) return false; // False if the manager already exists?
+  if (user) return false;
   await usersService.createNewUser(initialUsername, initialPassword);
-  return true; // Returns true if a new manager was created?
+  return true;
 }
 
 module.exports = {
--- a/cg/src/domains/management/users/users.helpers.js
+++ b/cg/src/domains/management/users/users.helpers.js
@@ -3,14 +3,14 @@ const UsersService = require('./users.service');
 const usersService = new UsersService();
 
 const initialUsername = 'admin';
-const initailPassword = 'clearGDPR';
+const initialPassword = 'clearGDPR';
 
 async function ensureInitialManagementUser() {
   // not sure if the initial user creds should be hardcoded or env variables... both feel bad
   const user = await usersService.getUserByUsername('admin');
-  if (user) return false;
-  await usersService.createNewUser(initialUsername, initailPassword);
-  return true;
+  if (user) return false; // False if the manager already exists?
+  await usersService.createNewUser(initialUsername, initialPassword);
+  return true; // Returns true if a new manager was created?
 }
 
 module.exports = {
--- a/cg/src/domains/management/users/users.service.js
+++ b/cg/src/domains/management/users/users.service.js
@@ -31,23 +31,27 @@ class UserService {
   async verifyUser(username, password) {
     const user = await this.getUserByUsername(username);
     if (!user) throw new Unauthorized('Invalid credentials');
-
     const passwordCheck = await this._comparePassword(password, user.password_hash);
-
     if (!passwordCheck) throw new Unauthorized('Invalid credentials');
-
     return { id: user.id };
   }
 
   async updatePassword(userId, newPassword) {
     const [user] = await this.db('users').where({ id: userId });
-
     if (!user) throw new BadRequest('User not found');
     await this.db('users')
       .update({ password_hash: await this._hashPassword(newPassword) })
       .where({ id: userId });
 
     return true;
+  }
+
+  async listManagementUsers() {
+    const managementUsers = await this.db('users')
+      .select('id')
+      .select('username');
+
+    return managementUsers;
   }
 }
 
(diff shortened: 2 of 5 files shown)


cg/test/management/users.api.spec.js
```diff
@@ -154,3 +154,70 @@ describe('User password update', () => {
     expect((await res2.json()).jwt).toBeTruthy();
   });
 });
+
+describe('Listing management users', () => {
+  it('Should display an empty list when no manager is registered', async () => {
+    //Given
+    await db('users').del(); // It's fine to delete the contents of the managers table because no other table references it
```
maksymilian-majer: it will be ok, because of the order Jest runs tests.
Nice gist: https://gist.github.com/tamlyn/dd6d54cdbe6eccd51a4bc61f5844bec4

cg/test/management/users.api.spec.js
```diff
@@ -154,3 +154,70 @@ describe('User password update', () => {
...
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${managementToken}`
+      }
+    });
+
+    //Then
+    expect(res.ok).toBeTruthy();
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual(expect.arrayContaining([]));
```
maksymilian-majer: This is wrong. It will pass even if the array isn't empty.
It should be:
`expect(await res.json()).toEqual([]);`

see this: https://repl.it/@maksymilian_maj/jest-arrays

cg/test/management/users.api.spec.js
```diff
@@ -154,3 +154,70 @@ describe('User password update', () => {
...
+  });
+
+  it('Should display correctly the registered managers', async () => {
+    //Given
+    await db('users').del(); // It's fine to delete the contents of the managers table because no other table references it
+    const managementToken1 = await managementJWT.sign({ id: '1' });
+    const managementToken2 = await managementJWT.sign({ id: '2' });
```
maksymilian-majer: You don't need a second token, `managementToken1` can be reused for both requests. It's basically a case when single manager with ID 1 creates two users. What you implemented is manager 1 and manager 2 are creating 1 user each.

cg/test/management/users.api.spec.js
```diff
@@ -154,3 +154,70 @@ describe('User password update', () => {
...
+      method: 'POST',
+      headers: { Authorization: `Bearer ${managementToken1}` },
+      body: {
+        username: 'manager1',
+        password: 'manager1_password'
+      }
+    });
+    await fetch('/api/management/users/register', {
+      method: 'POST',
+      headers: { Authorization: `Bearer ${managementToken2}` },
```
maksymilian-majer: use `managementToken1`

cg/test/management/users.api.spec.js
```diff
@@ -154,3 +154,70 @@ describe('User password update', () => {
...
+      }
+    });
+
+    //Then
+    expect(res.ok).toBeTruthy();
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          id: 5, // ID is 5 because 1 through 4 were deleted previously in the tests
```
maksymilian-majer: The ID's here are non deterministic. Unless you use the knex to fill the DB other tests can affect the ID numbering and break this test. Since you went for the route of calling register endpoint you should not care what IDs are being created and therefore you should change to:

```
id: expect.any(Number)
```

cg/src/domains/management/users/users.helpers.js
```diff
@@ -3,14 +3,14 @@ const UsersService = require('./users.service');
...
-const initailPassword = 'clearGDPR';
+const initialPassword = 'clearGDPR';
 
 async function ensureInitialManagementUser() {
   // not sure if the initial user creds should be hardcoded or env variables... both feel bad
   const user = await usersService.getUserByUsername('admin');
-  if (user) return false;
-  await usersService.createNewUser(initialUsername, initailPassword);
-  return true;
+  if (user) return false; // False if the manager already exists?
```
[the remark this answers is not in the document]
sindelio: Ok

cg/src/domains/management/users/users.helpers.js
```diff
@@ -3,14 +3,14 @@ const UsersService = require('./users.service');
...
+  await usersService.createNewUser(initialUsername, initialPassword);
+  return true; // Returns true if a new manager was created?
```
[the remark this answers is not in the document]
sindelio: Ok

cg/src/domains/management/users/users.service.js
```diff
@@ -31,24 +31,28 @@ class UserService {
...
     await this.db('users')
       .update({ password_hash: await this._hashPassword(newPassword) })
       .where({ id: userId });
 
     return true;
   }
+
+  async listManagementUsers() {
+    const managementUsers = await this.db('users')
+      .select('id')
```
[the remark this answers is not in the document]
sindelio: Hmm, I didn't know that

cg/test/management/users.api.spec.js
sindelio: Oks

cg/test/management/users.api.spec.js
sindelio: Ok

cg/test/management/users.api.spec.js
sindelio: Oks, what are other possibilities for this?

cg/test/management/users.api.spec.js
sindelio: We also could delete the 'users' table beforeEach test. It's good to have a clean state for each individual test

Jakedavies: it adds a lot of overhead, we should be able to get away with 1 db per file

PR #149, opened 2018-06-14
Added private routes support and login + logout

### LINKED ISSUE #133: Base management UI available to authenticated users
- [x] Protect the dashboard from unauthorized access
- [x] Add redirection to correct URL after login
- [x] Fix storing JWT in cookie
- [x] report bug with 500 error on wrong payload
- [x] support logout

We have the layout and components for the menu, so we should use that, but we should hook up routing and routes that will be protect from anonymous users.

maksymilian-majer:
Closes #133

### DIFF
--- a/admin/src/components/core/cards/dashboard/Sidenav.js
+++ b/admin/src/components/core/cards/dashboard/Sidenav.js
@@ -1,11 +1,26 @@
 import React from 'react';
-import { NavLink } from 'react-router-dom';
+import { NavLink, withRouter } from 'react-router-dom';
+import PropTypes from 'prop-types';
+
 import Icon from './Icon';
 import IconOverview from './../../../../assets/icons/overview.svg';
 import IconProcessors from './../../../../assets/icons/processors.svg';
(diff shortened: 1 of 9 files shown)


admin/src/components/core/cards/dashboard/Sidenav.js
```diff
@@ -1,11 +1,26 @@
...
+
 import Icon from './Icon';
 import IconOverview from './../../../../assets/icons/overview.svg';
 import IconProcessors from './../../../../assets/icons/processors.svg';
 import dashLogo from './../../../../assets/images/dash-logo.svg';
 
+import session from '../../../../helpers/Session';
+
 const Sidenav = props => {
+  // TODO: move when refactoring dashboard to presentational and container components
```
maksymilian-majer: This is temporary until we do the presentational/container components refactoring

Jakedavies:
Are we adding tests separately? It might be nice to make sure the privateRoute works properly and couple other high risk things

PR #154, opened 2018-06-14
Add validators for management endpoints and small refactors in the tests

### LINKED ISSUE #146: HTTP 500 error on wrong payload during login
**Steps to reproduce:**
Make a request to `/api/management/users/login` using an incorrect payload (ex. `email` instead of `username`

**Expected result:**
HTTP Error 400 (Bad Request)

**Actual result:**
HTTP Error 500 (Internal Server Error)

sindelio:
Tested in Postman and using the Jest test suite, all working.
I left some questions in pieces of the code. Also waiting for more input about the snapshoting before a full update of the tests

Closes #146

### DIFF
--- a/cg/src/domains/management/contract/contract.controller.js
+++ b/cg/src/domains/management/contract/contract.controller.js
@@ -1,4 +1,3 @@
-const { BadRequest } = require('../../../utils/errors');
 const ContractService = require('./contract.service');
 const { NotFound } = require('../../../utils/errors');
 
@@ -8,9 +7,6 @@ class ContractController {
   }
 
   async deployContract(req, res) {
-    if (!req.body.abiJson || !req.body.compiledData) {
-      throw new BadRequest(`The contract must contain values for 'abiJson' and 'compiledData'.`); // Never executed
-    }
     let parsedAbi = JSON.parse(req.body.abiJson);
     res.json({
       address: await this.service.deployContract(parsedAbi, req.body.compiledData)
--- a/cg/src/domains/management/contract/contract.controller.js
+++ b/cg/src/domains/management/contract/contract.controller.js
@@ -9,7 +9,7 @@ class ContractController {
 
   async deployContract(req, res) {
     if (!req.body.abiJson || !req.body.compiledData) {
-      throw new BadRequest(`The contract must contain values for 'abiJson' and 'compiledData'.`);
+      throw new BadRequest(`The contract must contain values for 'abiJson' and 'compiledData'.`); // Never executed
     }
     let parsedAbi = JSON.parse(req.body.abiJson);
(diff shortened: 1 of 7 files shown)


cg/src/domains/management/contract/contract.controller.js
```diff
@@ -9,7 +9,7 @@ class ContractController {
 
   async deployContract(req, res) {
     if (!req.body.abiJson || !req.body.compiledData) {
-      throw new BadRequest(`The contract must contain values for 'abiJson' and 'compiledData'.`);
+      throw new BadRequest(`The contract must contain values for 'abiJson' and 'compiledData'.`); // Never executed
```
sindelio: This error will never happen because of the Joi validator. Should we keep it anyway?

cg/src/domains/management/users/users.controller.js
```diff
@@ -10,14 +10,17 @@ class UsersController {
   async register(req, res) {
     const { username, password } = req.body;
     if (!username || !password) {
-      throw new BadRequest('Cannot register without an username or password');
+      throw new BadRequest('Cannot register without an username or password'); // This will never be called because we have a Joi validator
     }
```
sindelio: Same here

cg/src/domains/management/users/users.controller.js
```diff
@@ -10,14 +10,17 @@ class UsersController {
...
     await this.usersService.createNewUser(username, password);
     res.json({ success: true });
   }
 
   async login(req, res) {
     const { username, password } = req.body;
+    if (!username || !password) {
+      throw new BadRequest('Cannot login without an username or password'); // This will never be called because we have a Joi validator
+    }
```
sindelio: And here

cg/src/domains/management/users/users.controller.js
```diff
@@ -26,7 +29,7 @@ class UsersController {
   async updatePassword(req, res) {
     const { userId } = req.params;
     const { password } = req.body;
-    if (!password) throw new BadRequest('No password provided');
+    if (!password) throw new BadRequest('No password provided'); // Never executed
```
sindelio: And here

cg/src/domains/management/contract/contract.controller.js
Jakedavies: yes if its made redundant by the validator we can probably remove it

maksymilian-majer: Yes, remove all the validations that are handled by Joi.

cg/src/domains/management/users/users.controller.js
maksymilian-majer: Yes, remove

PR #156, opened 2018-06-14
added jwt token expiry and test

### LINKED ISSUE #144: Handle token expiration with JWTs
Right now we create JWTs that are permanent. We should either implement some store for sessions to make sure, that a token hasn't been expired or destroyed or we should add expiration (`exp`) to the token and check it when validating tokens. We need to handle it in subjects and management APIs. We should not do it for processors for now, but there should also be some more elaborate protection there later on (follow-up issue?).

Jakedavies:
closes #144

### DIFF
--- a/cg/test/subjects/subjects.api.spec.js
+++ b/cg/test/subjects/subjects.api.spec.js
@@ -1,4 +1,5 @@
 jest.mock('../../src/utils/blockchain/web3-provider-factory');
+jest.mock('../../src/utils/helpers');
 
 const encryption = require('./../../src/utils/encryption');
 const winston = require('winston');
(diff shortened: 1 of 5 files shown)


Jakedavies:
Tests appear to be failing because travis doesn't have the new env. Need to check with maks on how the test env is setup.

cg/test/subjects/subjects.api.spec.js
```diff
@@ -91,9 +91,27 @@ describe('Tests of subject giving consent', () => {
     );
   });
 
+  it('Should return an error if the jwt token is expired', async done => {
+    const token = await subjectJWT.sign({ test: '10' }, { expiresIn: 1 });
```
sindelio: test: '10' means ?
expiresIn: 1 is one second or millisecond?

cg/test/subjects/subjects.api.spec.js
```diff
@@ -91,9 +91,27 @@ describe('Tests of subject giving consent', () => {
...
+
+    expect.assertions(2);
+
+    setTimeout(async () => {
```
sindelio: waiting 1,1 seconds right

Jakedavies: yeah timeout is 1100 ms

cg/test/subjects/subjects.api.spec.js
Jakedavies: test:10 means nothing, just a payload 
expiresIn1 is 1 second yes

PR #163, opened 2018-06-18
Update smart contract to support right to rectification

### LINKED ISSUE #157: Record the update event when a users request to update their data is approved

sindelio:
Need more tests though
Closes #157


cg/smart_contracts/ClearGDPR_commented.sol
```diff
@@ -1,131 +1,75 @@
-/*  ClearGDPR smart contract version 5
+/*  ClearGDPR smart contract version 6
 *       A piece of art made by CleverTech
 *       Author contact: sindelio.lima@clevertech.biz
```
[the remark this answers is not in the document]
sindelio: Ok

sindelio:
All tested and working.
It's ready to receive reviews!

cg/scripts/blockchain-function-test.js
```diff
@@ -0,0 +1,32 @@
+const {
```
sindelio: This test script is good to test the new smart contract functions inside the system, so let's leave it here now with a nicer name ;)

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,17 +1,26 @@
-pragma solidity 0.4.21; 
+/*  ClearGDPR smart contract version 6
```
sindelio: Now we only have one version of the smart contract, with the most important remarks and comments

PR #167, opened 2018-06-18
Feature/admin list users

### LINKED ISSUE #152: Setup front-end tests on Travis
- [x] run front-end tests in the testing phase
- [ ] check admin warnings in testing phase
- [ ] re-enable treating warnings as errors when building admin in deploy phase

We have temporarily disabled warnings on Travis, so that dist builds go through. In this task we need to enable them back and run the tests as part of build process, so the builds are properly approved.

Blocked by #120 and #141

### LINKED ISSUE #162: Listing users in admin
As a manager when I go to the `/users` route I want to see all management users that have been registered in the system.

For each user, I want to have a button that allows me to change a password. Changing password feature will be a separate issue

maksymilian-majer:
Closes #162, closes #152 

Adds minor fixes (button on Login screen had wrong class) and example tests for container state/rendering. Some refactoring to keep APIs of components more generic.

### DIFF
--- a/admin/src/containers/Users/Users.js
+++ b/admin/src/containers/Users/Users.js
@@ -5,7 +5,8 @@ import session from '../../helpers/Session';
 
 class UsersContainer extends React.Component {
   state = {
-    users: []
+    users: [],
+    isLoading: true
   };
 
   async getUsers() {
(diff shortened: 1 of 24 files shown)


admin/src/containers/Users/Users.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import Users from '../../components/Users/Users';
+import config from '../../config';
+import session from '../../helpers/Session';
+
+class UsersContainer extends React.Component {
```
Jakedavies: do we have any loading state functionality? It would be nice to have some sort of indication of loading for this stuff in state.

maksymilian-majer: @Jakedavies we do have a loader and I added it. I forgot about it..

PR #173, opened 2018-06-19
Feature/admin panel context

### LINKED ISSUE #159: Handle interaction with side panel from child components
We should have a way to open side panel in `MainLayout` component from children. Since we're not using Redux and a shared store, then an option can be to create a HOC like `withRouter` that will provide an interface to interact with the panel (i.e. open it and set its content).

The components rendered in the panel also need a way to close it.

maksymilian-majer:
Closes #159

### DIFF
--- a/admin/src/containers/MainLayout/PanelContext.js
+++ b/admin/src/containers/MainLayout/PanelContext.js
@@ -1,0 +1,53 @@
+import React, { Component, createContext } from 'react';
+import PropTypes from 'prop-types';
+
+const PanelContext = createContext({
+  component: null,
+  title: null,
+  props: {},
+  openPanel: () => {},
+  closePanel: () => {},
+  isPanelOpen: false
+});
+
+export class PanelProvider extends Component {
+  static propTypes = {
+    children: PropTypes.oneOf(
+      PropTypes.element,
+      PropTypes.object,
+      PropTypes.arrayOf(PropTypes.element)
+    )
+  };
+
+  openPanel = (component, title = null, props = {}) => {
+    this.setState({
+      component,
+      title,
+      isPanelOpen: true,
+      props
+    });
+  };
+
+  closePanel = () =>
+    this.setState({
+      component: null,
+      title: null,
+      isPanelOpen: false,
+      props: {}
+    });
+
+  state = {
+    component: null,
+    title: null,
+    props: {},
+    openPanel: this.openPanel,
+    closePanel: this.closePanel,
+    isPanelOpen: false
+  };
+
+  render() {
+    return <PanelContext.Provider value={this.state}>{this.props.children}</PanelContext.Provider>;
+  }
+}
+
+export const PanelConsumer = PanelContext.Consumer;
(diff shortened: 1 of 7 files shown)


admin/src/containers/MainLayout/PanelContext.js
```diff
@@ -0,0 +1,53 @@
...
+export class PanelProvider extends Component {
+  static propTypes = {
+    children: PropTypes.oneOfType([
+      PropTypes.element,
+      PropTypes.object,
+      PropTypes.arrayOf(PropTypes.element)
+    ])
+  };
+
+  openPanel = (component, title = null, props = {}) => {
```
fjcero: Why not just `openPanel() {}` ?

admin/src/containers/MainLayout/PanelContext.js
```diff
@@ -0,0 +1,53 @@
...
+    this.setState({
+      component,
+      title,
+      isPanelOpen: true,
+      props
+    });
+  };
+
+  closePanel = () =>
```
fjcero: Same than above!

admin/src/containers/MainLayout/PanelContext.js
```diff
@@ -0,0 +1,53 @@
...
+    this.setState({
+      component: null,
+      title: null,
+      isPanelOpen: false,
+      props: {}
+    });
+
+  state = {
```
fjcero: I usually put default state on the top of the component declaration

admin/src/containers/MainLayout/PanelContext.js
juanmnl: This syntax ensures `this` is bound within the fn() but it's also experimental. It's supposed to be safe with babel, but I use it on prototypes :P It can be safer to fn() {} and bind `this` in the constructor() :D

admin/src/containers/MainLayout/PanelContext.js
maksymilian-majer: I took this idea from an article:
https://medium.com/@BogdanSoare/how-to-use-reacts-new-context-api-to-easily-manage-modals-2ae45c7def81

screen shot 2018-06-20 at 11 29 40

I haven't tested it with state at the top, but it is likely to break.

admin/src/containers/MainLayout/PanelContext.js
maksymilian-majer: I don't understand the question? The `PanelProvider` needs to provide an `openPanel` function to the components that want to interact with this context. I'm setting the state so that the component gets re-rendered. That is how we can simulate something similar to actions and reducers that change the central store similarly to redux. Let's discuss this today and I'll refactor in a separate task if you have something better in mind.

admin/src/containers/MainLayout/PanelContext.js
maksymilian-majer: Oh, now I understand. We can experiment without assigning the function to a property. I'll play with this in a next iteration.

Issue #182, opened 2018-06-21
Form component


fjcero:
`react-form` is a way to abstract form control, is not hardly opinionated but it needs more work to allow reusable components and facilitate implementation.

Some of the key points are:

- Handle Errors, for now global, but will be nicer to allow inline errors as well
- Styling should be applied by default
- A schema could be specified (probably a combination with `react-jsonschema-form`)

Some interesting reads on how do forms without having to use redux
- https://moduscreate.com/blog/reactjs-form-validation-approaches/
- https://medium.com/javascript-inside/some-thoughts-on-forms-in-react-9ca2d9078c20

Libs proposed/used:
- https://github.com/mozilla-services/react-jsonschema-form
- https://github.com/react-tools/react-form

maksymilian-majer:
@fjcero so far I've used the forms from @juanmnl for issue #174 which is in progress, so let's refactor later

fjcero:
@maksymilian-majer `react-form` was renamed to `informed` -> https://joepuzzo.github.io/informed

Either way, no major changes are required to use this library, and I'm researching a way to test forms in a more sustainable way instead of repeating code by each form implementation. Probably abstracting the generic Form in a `common` component will be a good idea. I will like to chat with you before working further in this task

maksymilian-majer:
I see we naturally went to use react-forms and I created an issue to migrate to informed (#209)
Can this be closed down?

fjcero:
Yeah, let’s continue the migration in #209.

PR #183, opened 2018-06-21
Login form improvements

### LINKED ISSUE #142: Form validation  on login

### LINKED ISSUE #180: Add a loader when logging in
We have a loader when loading users. You can use the same approach.

fjcero:
Close #142
Close #180

- Added support to forms with `react-forms`. This is a step forward to implement more reusable form components.
- Added loader when the form is submitted.
- Wrapped parent `handleLogin` into a promise to pass down to components.
- Added Errors wrapper, should be abstracted in Form Component (WIP) (#182)
- Some fixes to prop-types

### DIFF
--- a/admin/src/containers/Login/Login.js
+++ b/admin/src/containers/Login/Login.js
@@ -11,30 +11,24 @@ export class LoginContainer extends React.Component {
     const { history, location } = this.props;
     const { from } = location.state || { from: { pathname: '/' } };
 
-    return new Promise((resolve, reject) => {
-      fetch(`${config.API_URL}/api/management/users/login`, {
-        method: 'POST',
-        body: JSON.stringify({
-          username,
-          password
-        }),
-        headers: {
-          'Content-Type': 'application/json'
-        }
-      })
-        .then(async res => {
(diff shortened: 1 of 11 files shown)


admin/src/containers/Login/Login.js
```diff
@@ -6,35 +6,39 @@ import Login from '../../components/Login/Login';
...
-        password
-      }),
-      headers: {
-        'Content-Type': 'application/json'
-      }
-    })
-      .then(async res => {
-        session.set(await res.json());
-        history.push(from);
+    return new Promise((resolve, reject) => {
```
maksymilian-majer: Fetch already returns a promise so I suggest we return it instead of creating a new one. We risk getting into the [forgotten promise antipattern](http://taoofcode.net/promise-anti-patterns/) – not that the code will not work.

Jakedavies: yeah there is no reason to create a new promise here, the fetch one is better 👍

Jakedavies:
looks fine as long as we address that funny promise issue.

PR #190, opened 2018-06-22
Remove managers endpoint

### LINKED ISSUE #171: Removing management users in cg
We can register the users, but there's no way of removing them. I think there are no relations to users table so no soft delete required, just delete from DB.

sindelio:
Closes #171 
Tests and Postman working.


cg/src/domains/management/users/users.service.js
```diff
@@ -21,13 +21,28 @@ class UsersService {
     return user;
   }
 
+  async getUserById(userId) {
+    const [user] = await this.db('users')
+      .where({ id: userId })
+      .limit(1);
```
maksymilian-majer: There's a unique index on id, so the `limit` is not required, but it is ok.

PR #192, opened 2018-06-25
Feature/apply rectification update on approval

Jakedavies:


cg/test/management/subjects.api.spec.js
```diff
@@ -599,9 +608,43 @@ describe('List subjects that have given consent', () => {
       expect(res.status).toEqual(200);
 
       const [request] = await db('rectification_requests').where({ id: rectificationRequestId });
+      expect(request.status).toEqual('APPROVED');
+    });
+
+    it('Should apply the rectification if the request is approved', async () => {
```
sindelio: There's no negative cases here?
What happens if one tries to approve a request that wasn't made for example

Jakedavies: There is a negative case right under your comment

Jakedavies: All of the negative cases were handled by the previous PR where the endpoint was implemented, this was adding additional functionality for propagating the update specifically. I didn't think of any failure cases for that part outside of what is already there.

PR #201, opened 2018-06-27
Feature/handle jwt expire on frontend

### LINKED ISSUE #145: Handle token expiration gracefully on the front-end
When #144 is implemented we should detect that on the front-end and redirect to login page.
We should handle this both for management and demo front-end.

Jakedavies:
closes #145

### DIFF
--- a/admin/src/helpers/internal-fetch.js
+++ b/admin/src/helpers/internal-fetch.js
@@ -1,5 +1,5 @@
-import merge from 'lodash/object/merge';
-import session from '../../helpers/Session';
+import { merge } from 'lodash/object';
+import session from './Session';
 
 const internalFetch = (url, config = {}) => {
   const internalConfig = merge(
@@ -15,6 +15,8 @@ const internalFetch = (url, config = {}) => {
   return fetch(url, internalConfig).then(async res => {
     if (res.status === 400 && (await res.json()).error === 'JWT token expired') {
       session.destroy();
+      window.location.assign('/login?expired=1');
+      // throw new Error('Session expired');
     } else {
       return res;
     }
--- a/admin/src/helpers/internal-fetch.js
+++ b/admin/src/helpers/internal-fetch.js
@@ -1,0 +1,24 @@
+import merge from 'lodash/object/merge';
+import session from '../../helpers/Session';
+
+const internalFetch = (url, config = {}) => {
+  const internalConfig = merge(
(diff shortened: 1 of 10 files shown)


admin/src/helpers/internal-fetch.js
```diff
@@ -0,0 +1,26 @@
+import { merge } from 'lodash/object';
+import session from './Session';
+
+const internalFetch = (url, config = {}) => {
```
maksymilian-majer: @Jakedavies after you rebase can you refactor to use internalFetch on the changes I submitted?

admin/src/helpers/internal-fetch.spec.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    const res = await internalFetch('');
+    expect(res.ok).toEqual(true);
+    expect(res.status).toEqual(200);
+    expect(await res.json()).toEqual({ test: 'test' });
+  });
+
+  it('Should redirect and destroy the session if JWT is expired', async () => {
+    session.getToken = jest.fn().mockImplementationOnce(() => 'token');
```
maksymilian-majer: I'm mocking session using `jest.mock('../../helpers/Session');` in other places, but it's ok too :)

cg/src/utils/jwt.js
```diff
@@ -42,7 +42,7 @@ exports.managementJWT = new JWT(
   'HS256',
   process.env.MANAGEMENT_SECRET,
   process.env.MANAGEMENT_SECRET,
-  { expiresIn: getTokenExpiry() }
+  { expiresIn: 5 }
```
maksymilian-majer: I think you intended this for testing purposes only and should be reverted back to `getTokenExpiry()`?

fjcero: Can we define an env file for testing or doesn’t make sense?

maksymilian-majer: I believe there is a default for token expiry, so nothing needs to be done for this setting

admin/src/helpers/internal-fetch.js
Jakedavies: will do

cg/src/utils/jwt.js
Jakedavies: i did not mean to change this. will revert

PR #202, opened 2018-06-27
Processors list

### LINKED ISSUE #161: Listing processors
As a manager when I go to the `/processors` route I want to see all the processors that have been added to the system.

For each processor, I want to have a button that allows me to edit and delete them. These features will have separate issues.

fjcero:
Closes #161 

- Test helpers separated in a different directory inside `/src` (just for avoid struggling with import issue)
- Added processors data factory as test helpers
- Added container to Main layout
- Added tests


admin/src/tests/helpers/TestUtils.js
```diff
@@ -0,0 +1,4 @@
+/**
+ * Flush all pending resolved promise handlers.
+ */
+export const flushPromises = () => new Promise(resolve => setImmediate(resolve));
```
maksymilian-majer: @fjcero did you notice I also added a helper and it's already merged to master. Can you rebase and make sure we only have one TestUtils?

PR #203, opened 2018-06-27
Processors form

### LINKED ISSUE #204: Form to Manage processors (edit)
A form to manage processors details should be provided. Once is done, the same form should be used for creation. Considering that, the form should be reusable and should be well tested.

fjcero:
**Not ready to be merged**

Closes #204 

- Some concerns on Form testing should be addressed
- Processors Form for editing

### DIFF
--- a/admin/src/components/Processors/EditProcessor.js
+++ b/admin/src/components/Processors/EditProcessor.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form, Text, TextArea, Checkbox } from 'react-form';
+
+import Loader from '../core/cards/dashboard/Loader';
+
+export const EditProcessor = props => {
+  const scopes = {
+    'Full Name': 'user.fullName',
+    Email: 'user.email',
+    'Phone Number': 'user.phoneNumber'
+  };
+
+  const renderForm = () => (
+    <React.Fragment>
+      <label htmlFor="name">Name:</label>
+      <Text field="name" id="name" />
+
+      <label htmlFor="description">Description:</label>
+      <TextArea field="description" id="description" />
+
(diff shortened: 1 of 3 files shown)


admin/src/components/Processors/EditProcessor.js
```diff
@@ -0,0 +1,62 @@
...
+  const scopes = {
+    'Full Name': 'user.fullName',
+    Email: 'user.email',
+    'Phone Number': 'user.phoneNumber'
+  };
+
+  const renderForm = () => (
+    <React.Fragment>
+      <label htmlFor="name">Name:</label>
+      <Text field="name" id="name" />
```
maksymilian-majer: Have you seen how I used our `TextInput` control and validation logic in a container?
Maybe we should consider using that approach – I think it will cut down on the code a little bit and make it consistent across different forms. Let's chat about it after standup.

fjcero: Yep! I saw that, just flagged the PR as WIP, to check with you how we can avoid repeating the part of code related to events on form controls, perhaps you have some ideas after making the user forms. Let discuss this a little bit!

admin/src/components/Processors/EditProcessor.js
```diff
@@ -0,0 +1,62 @@
...
+      <label>Scopes:</label>
+      {Object.keys(scopes).map((scope, i) => <Checkbox key={i} field={`scope[${scope}]`} />)}
+
+      <button type="submit" className="btn">
+        Submit
+      </button>
+    </React.Fragment>
+  );
+
+  return <form onSubmit={props.onSubmit}>{props.isLoading ? <Loader /> : renderForm()}</form>;
```
Jakedavies: form -> Form

PR #216, opened 2018-06-27
Added proptypes to kitchen sink

### LINKED ISSUE #141: Add proptypes to all components emitting warnings
Most kitchen sink cards are emitting warnings on prop-types. Some of these cards are already used as official presentational components, and some work is needed to improve support for testing and validations.

fjcero:
Closes #141 

To remove warnings. Also, some proptypes could be standardized I will create a task for that.


fjcero:
@maksymilian-majer It removes all the warnings, all components have prop-types so far.

PR #220, opened 2018-06-27
Added HOCs to wrap elements in context providers

### LINKED ISSUE #219: Provide contexts to the main layout in a nicer way
We will need to wrap the MainLayout in many contexts, so ideally we'll have a HOC that makes it nicer and more maintainable.

maksymilian-majer:
Closes #219

### DIFF
--- a/admin/src/containers/MainLayout/MainLayout.js
+++ b/admin/src/containers/MainLayout/MainLayout.js
@@ -1,5 +1,7 @@
 import React, { Component } from 'react';
 import { Route, Switch } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import lodash from 'lodash';
 
 import { PanelProvider, PanelConsumer } from './PanelContext';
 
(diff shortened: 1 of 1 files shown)


admin/src/containers/MainLayout/MainLayout.js
```diff
@@ -38,42 +40,51 @@ class MainLayoutContainer extends Component {
...
+      <Component {...props}>{props.children}</Component>
+    </ContextProvider>
+  );
+  wrapper.propTypes = {
+    children: PropTypes.node
+  };
+  return wrapper;
+};
+
+const withProviders = lodash.flow.apply(null, [
```
maksymilian-majer: using apply instead of just calling `lodash.flow()` is a workaround for some prettier / eslint issue that would trigger a pre-commit hook which would then fail to compile. Super awkward, maybe we can fix this together tomorrow.

Issue #222, opened 2018-06-28
Warnings on the smart contract about line size and indentation


sindelio:
The most used Visual Studio Code extension for Solidity (by far) is giving me warnings about the maximum amount of characters per line, which should be less than 145, and the indentation, which should be 4 spaces. I don't think this will cause any issues but just to play it safe it will be changed to appease the extension

PR #224, opened 2018-06-28
Fix/controllers not treated as processors in smart contract

### LINKED ISSUE #221: Controller not treated as a processor properly in the smart contract
Until now the controller was not being treated as a processor properly in the smart contract, and the back-end was doing that as a workaround (i.e. calling setProcessors from blockchain.js with the controller address in the middle of the processors addresses). The fact that the controller is another processor should be transparent to the back-end

sindelio:
Closes #221, #222 
Work in progress, 10 tests are not passing yet


sindelio:
All tests are passing, the smart contract functions were tested in Remix and inside the system with the test script. It's ready for reviews o/

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,153 +1,163 @@
...
-  }
+    function recordConsentGivenTo(bytes32 _subjectId, address[] _processorsConsented) public onlyController notErased(_subjectId) returns(bool){
+        require(areAllValidProcessors(_processorsConsented));
+        for(uint256 i = 0; i < processors.length; i++){
+            subjects[_subjectId].statePerProcessor[processors[i]] = State.notShareable;  
+        }
+        for(i = 0; i < _processorsConsented.length; i++){
+            subjects[_subjectId].statePerProcessor[_processorsConsented[i]] = State.shareable;  
+        }
+        subjects[_subjectId].statePerProcessor[controller] = State.shareable;
```
sindelio: the controller always has consent. I think the state shareable is not the best name for when a Quorum node has consent. We should probably rename it, because the controller can't actually share the subject's data without the subject's consent for a specific processor. Suggestion: unconsented, consented, erased, objected

cg/scripts/blockchain-function-test.js
```diff
@@ -1,27 +1,89 @@
...
-  getPastEvents
-} = require('../src/utils/blockchain');
-
-getRectificationCount('0x0000000000000000000000000000000000000000000000000000000000000001')
-  .then(console.log)
-  .catch(console.error)
-  .then(() => process.exit(0));
-
-// recordRectificationByController('0x0000000000000000000000000000000000000000000000000000000000000001')
+// docker-compose exec cg node scripts/blockchain-function-test.js
```
sindelio: The test script is complete with the function calls now. This is good to know how to call each function and test it

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,153 +1,163 @@
-/*  ClearGDPR smart contract version 6
+/*  ClearGDPR smart contract version 7
 *       A piece of art made by CleverTech
 *       Author contact: sindelio.lima@clevertech.biz
 *       Coffee => Code!
+*
+*   IMPORTANT NOTES !!!!
```
sindelio: Please read these notes!

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,153 +1,163 @@
...
+        return true;
+    }
 
 }
 
 /*  ASSUMPTIONS:
+*       1. The controller is a processor, but processors are not controllers.
 *       1. Only 1 controller per smart contract.
 *       2. Users can only interact directly with the controller. That means data access requests don't need to be informed to the processors.
 *       3. Only controllers can initiate data erasure events.
```
sindelio: What if a processor want to delete his copy of the subject's data?

cg/scripts/blockchain-function-test.js
maksymilian-majer: Yeah, I think it makes sense to keep this.

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,135 +1,152 @@
...
-  
-  function isProcessor(address _processor) public view returns(bool){ //0x1 for remix, since it's the only argument
-    for(uint256 i = 0; i < processors.length; i++){
-      if(processors[i] == _processor){
+    function setProcessors(address[] _newProcessors) public returns(bool){
+        processors = new address[](_newProcessors.length + 1);
+        processors[0] = controller;
+        for(uint256 i = 0; i < _newProcessors.length; i++){
+            require(_newProcessors[i] != controller);
+            for(uint256 j = i + 1; j < _newProcessors.length; j++){ // Check if the processors are unique
```
sindelio: Now the processors have to be unique for the execution to complete, else the function will revert to the state before it was called

cg/test/subjects/subjects.api.spec.js
```diff
@@ -146,12 +146,14 @@ describe('Tests of subject giving consent', () => {
         Authorization: `Bearer ${token}`
       }
     });
-
+    // console.log(await res.text());
     // Then (Assert)
     let subjectIdHashed = hash('1');
+    //console.log(subjectIdHashed);
```
Jakedavies: remove these commented console.logs

Jakedavies:
I _think_ this looks good.

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,135 +1,152 @@
...
+                require(_newProcessors[i] != _newProcessors[j]);
+            }
+            processors[i + 1] = _newProcessors[i];
+        }
         return true;
-      }
+    } 
+    //processors = _newProcessors; //Interesting line
```
maksymilian-majer: Why is this line interesting? Shouldn't it just go away? ;)

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,135 +1,152 @@
...
     }
-    return false;
-  }
   
-  function areAllValidProcessors(address[] _processors) public view returns(bool){
-    for(uint256 i = 0; i < _processors.length; i++){
-      require(isProcessor(_processors[i]));
+    function areAllValidProcessors(address[] _processors) public view returns(bool){
+        for(uint256 i = 0; i < _processors.length; i++){
+            require(isProcessor(_processors[i]));
```
maksymilian-majer: Does it automatically return false if require fails?

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,135 +1,152 @@
...
-    return true;
-  }
   
-  function recordProcessorsUpdate(address[] _newProcessors) public onlyController returns(bool){
-    require(setProcessors(_newProcessors));
-    emit Controller_ProcessorsUpdated(_newProcessors);  
-    return true;
-  }
+    function recordProcessorsUpdate(address[] _newProcessors) public onlyController returns(bool){
+        require(setProcessors(_newProcessors));
```
maksymilian-majer: Same question as above for `require`

cg/smart_contracts/ClearGDPR.sol
maksymilian-majer: I agree. Will require refactoring of consts in JS codebase.

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,135 +1,152 @@
...
-    emit Controller_SubjectDataAccessed(_subjectIdHash);
-    return states;
-  }
+    function recordAccessByController(bytes32 _subjectId) public onlyController notErased(_subjectId) returns(State[]){
+        State[] memory states = new State[](processors.length);
+        for(uint256 i = 0; i < processors.length; i++){
+            states[i] = subjects[_subjectId].statePerProcessor[processors[i]];
+        }
+        emit Controller_SubjectDataAccessed(_subjectId);
+        return states;
```
maksymilian-majer: Not sure what we need these states for. Won't this just return an array like:
`[ shareable, notsherable, erased, erase, ... ]`
???

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -1,135 +1,152 @@
...
+        return true;
+    }
 
-  function recordErasureByController(bytes32 _subjectIdHash) public onlyController notErased(_subjectIdHash) returns (bool){
-    subjects[_subjectIdHash].isErased = true;
-    emit Controller_SubjectDataErased(_subjectIdHash);
-    return true;
-  }
+    function recordErasureByController(bytes32 _subjectId) public onlyController notErased(_subjectId) returns(bool){
+        subjects[_subjectId].isErased = true;
```
maksymilian-majer: How about this line above?

```
subjects[_subjectId].statePerProcessor[controller] = State.shareable;
```

Don't we also need to set erased state in the `statePerProcessor` array?

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -139,15 +156,10 @@ contract ClearGDPR {
...
-*       2. Controllers and processors have a different set of possible actions in the system. Hence the events each can fire are also different.
-*      We need a way to enforce this difference in the back-end if we stick with this smart contract.
-*       3. There's a bug when we use bytes32 instead of address to identificate the processors. The memory seems to not be cleared out. 
-*       
 *   TODO: 
 *       1. Change accessibility of setters to private (they are public for testing purposes)
 *       2. Implement the add/remove function of processor nodes in the network, or use a system ID for each processor, which would be simpler
 *       3. Study more about memory allocation/deallocation in Solidity. Will need to dive in inline Assembly
 *       4. Consider that each processor can delete their data of a subject, without the data being deleted from the whole network
+*       5. When a subjects gives consent, does that functions as a rectification of data? If not, his data should not be altered when he gives consent again
```
maksymilian-majer: good point!

cg/test/blockchain-setup.js
```diff
@@ -5,11 +5,10 @@ const ContractService = require('../src/domains/management/contract/contract.ser
...
...6001805490508210156115885760026000866000191660001916815260200190815260200160002060020160006001848154811015156114d457fe5b9060005260206000200160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060009054906101000a900460ff16838381518110151561155557fe5b90602001906020020190600281111561156a57fe5b9081600281111561157757fe5b815250508180600101925050611498565b7f5cee2268933cd873c783419e5be7668e121517767fe669a5a703b09426b138158560405180826000191660001916815260200191505060405180910390a1829350505050919050565b6000809054906101000a900473ffffffffffffffffffffffffffffffffffffffff1681565b828054828255906000526020600020908101928215611670579160200282015b8281111561166f5782518260006101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff16021790555091602001919060010190611617565b5b50905061167d9190611681565b5090565b6116c191905b808211156116bd57600081816101000a81549073ffffffffffffffffffffffffffffffffffffffff021916905550600101611687565b5090565b905600a165627a7a723058200ad03be1f5d4bf702118989ff2ff8bc1cbb12b05d9e49a089fcbd6869a300a950029';
```
maksymilian-majer: Maybe add an issue to have this only in one place – very error prone the way it is now..

cg/test/management/contract.api.spec.js
```diff
@@ -160,15 +160,12 @@ describe('Event feed websocket', () => {
...
     socketSubscription.on('message', event => {
       const data = JSON.parse(event);
       expect(data.eventName).toEqual('Controller_ConsentGivenTo');
-      expect(data.params.subjectIdHash).toEqual(sha3('websocket-test'));
-      expect(data.params.newProcessorsWhiteListed).toHaveLength(1);
-      expect(data.params.newProcessorsWhiteListed[0].toLowerCase()).toEqual(
-        process.env.CONTRACT_OWNER_ADDRESS.toLowerCase()
-      );
+      expect(data.params.subjectId).toEqual(sha3('websocket-test'));
+      expect(data.params.processorsConsented).toHaveLength(0);
```
maksymilian-majer: Is it OK, that on `Controller_ConsentGivenTo` the `data.params.processorsConsented` array is empty?

cg/test/subjects/subjects.api.spec.js
```diff
@@ -146,12 +146,14 @@ describe('Tests of subject giving consent', () => {
         Authorization: `Bearer ${token}`
       }
     });
-
+    // console.log(await res.text());
```
maksymilian-majer: Please remove these commented out logs. Including all the below occurrences.

cg/test/subjects/subjects.api.spec.js
sindelio: k

cg/smart_contracts/ClearGDPR.sol
sindelio: It can go away, but it's to remember us that we can assign array pointers in Solidity if the arrays have the same length. But I will remove it, and leave it in my Solidity notes

cg/smart_contracts/ClearGDPR.sol
sindelio: No, it reverts anything that was made up to that point in the function, to the state prior to the function call. The return true is to quickly test if the function executed successfully, even if the behavior was not correct

cg/smart_contracts/ClearGDPR.sol
sindelio: This require checks if the processors were updated with success. It does not return false if it fails, the entire functions reverts and we get a revert exception in the Ethereum Virtual Machine

cg/smart_contracts/ClearGDPR.sol
sindelio: Yeah, we can work some more in this. The idea is to get the state of the subject's data for each processor it has given consent. A better response would be:

{
  processor1: notShareable
  processor2: erased
  ...
}

But since it's difficult to return JSON in Solidity, although I think it is possible, were returning only the states for each processor. A simple change would be return the array with the current processors in the system along with the array with its respective state. then the back-end could iterate over the arrays, creating a nice JSON for the front-end

cg/smart_contracts/ClearGDPR.sol
sindelio: No because so far that's a property of the subject's data. We still had not considered the case that a processor deletes his copy of the data, without a request to be forgotten being issued by the subject. We should create an issue to handle this case in the future

cg/smart_contracts/ClearGDPR.sol
sindelio: Probably anther issue to look in the future. Will create those

cg/test/blockchain-setup.js
sindelio: Agree

cg/test/management/contract.api.spec.js
sindelio: Actually it is not, because the controller should be in that array. If we call recordConsentGivenTo without any processors (with an empty array), it will still set the controller as consented. Will look into this

cg/test/subjects/subjects.api.spec.js
sindelio: k

cg/smart_contracts/ClearGDPR.sol
sindelio: In the right to restrict I will have a good setup to handle this

cg/test/management/contract.api.spec.js
sindelio: Created a task to handle this, this PR is too big

Issue #200, opened 2018-06-26
Login form errors are not being displayed


Jakedavies:
Something is swallowing them?

PR #226, opened 2018-06-28
Fix/login errors not displaying

### LINKED ISSUE #225: Refactor internalFetch to handle api errors nicely
@sindelio Please fill it.

**Additional context**

Jakedavies:
closes #225 #200 

I probably need to rebase and do a little bit of refactoring against master so the new internalFetch works for everything, but I think overall this nets us a lot less duplicated code

### DIFF
--- a/admin/src/containers/Users/UsersContext.js
+++ b/admin/src/containers/Users/UsersContext.js
@@ -1,6 +1,5 @@
 import React, { Component, createContext } from 'react';
 import PropTypes from 'prop-types';
-import session from '../../helpers/Session';
 import config from '../../config';
 import internalFetch from './../../helpers/internal-fetch';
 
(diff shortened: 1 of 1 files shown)


Jakedavies:
This also allows internalFetch to be used when the user isn't login (checks for token existence before attaching it)

Jakedavies:
An important thing to note here is that we parse the json body by default now with internalFetch, to save everyone parsing it themselves. So when consuming internalFetch you don't need to await res.json()

admin/src/containers/Users/ChangePassword.js
```diff
@@ -18,44 +18,18 @@ export class ChangePasswordContainer extends React.Component {
   };
 
   async changePassword(userId, password) {
-    const response = await internalFetch(
```
maksymilian-majer: Is it supposed to affect this file?
I think there's some merge gone wrong.

admin/src/containers/Users/Users.js
```diff
@@ -19,9 +19,7 @@ export class UsersContainer extends React.Component {
   };
 
   async getUsers() {
-    const response = await internalFetch(`${config.API_URL}/api/management/users/list`);
-
-    return await response.json();
+    return await internalFetch(`${config.API_URL}/api/management/users/list`);
```
maksymilian-majer: Same here. I think the new master is already using the context here.

PR #239, opened 2018-06-29
Feature/toasts

### LINKED ISSUE #195: Handling error/success notifications
- [x] password successfully change
- [x] user successfully registered
- [x] user successfully deleted
- [x] fetch errors in case of all unhandled errors in the above methods
- [x] deleting user (when #194 closed)

Sometimes there can be errors or success notifications that we could show to the user "detached" from the UI. Like after successful password change a nice popup would be welcome. 

We could use something like:
https://fkhadra.github.io/react-toastify/
https://tomchentw.github.io/react-toastr/
https://github.com/jesusoterogomez/react-notify-toast

Let's have the toasts show from top-center.

Jakedavies:
closes #195 

Addded a few extra ones than the ones we listed out, work well and seem to match the UI style without any restyling, although @juanmnl might want to take a look still.

### DIFF
--- a/admin/src/containers/App.js
+++ b/admin/src/containers/App.js
@@ -1,15 +1,14 @@
 import React, { Component } from 'react';
-import logo from 'assets/logo.png';
+import Routes from 'routes';
+import { ToastContainer } from 'react-toastify';
+import 'react-toastify/dist/ReactToastify.css';
 
 class App extends Component {
   render() {
     return (
       <div className="App">
-        <header className="App-header">
-          <img src={logo} className="App-logo" alt="logo" />
-          <h1 className="App-title">Welcome to React</h1>
-        </header>
-        <p className="App-intro">this is app boilerplate</p>
+        <Routes />
+        <ToastContainer />
       </div>
     );
   }
--- a/admin/src/containers/Login/Login.test.js
+++ b/admin/src/containers/Login/Login.test.js
@@ -4,6 +4,7 @@ import toJson from 'enzyme-to-json';
 import { createMemoryHistory } from 'history';
 import { flushPromises } from 'tests/helpers/TestUtils';
 import { LoginContainer } from './Login';
+import { toast } from 'react-toastify';
 
 jest.mock('helpers/Session');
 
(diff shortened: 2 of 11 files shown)


admin/src/containers/Login/Login.test.js
```diff
@@ -69,4 +70,24 @@ describe('(Container) Login', () => {
...
+        ok: true,
+        status: 200,
+        json: () =>
+          Promise.resolve({
+            token: 'token'
+          })
+      })
+    );
+
+    toast.success = jest.fn();
```
maksymilian-majer: why aren't you using `jest.mock('react-toastify')`?

admin/src/containers/Users/UsersContext.test.js
```diff
@@ -146,6 +147,27 @@ describe('UsersProvider', () => {
...
+
+    const { component } = setupShallow();
+
+    try {
+      await component.instance().deleteUser(4);
+    } catch (e) {
+      //not-empty
+    }
+
+    expect(toast.error).toHaveBeenCalledWith('An error occurred: Error message');
```
maksymilian-majer: I would also test for `toast.success` to not have been called. I kind of think there's something odd happening there when catching errors on deleting users. I think that you'd have to do `toast.error = jest.fn();` in `beforeEach` in such case.

admin/src/containers/App.js
```diff
@@ -1,15 +1,14 @@
 import React, { Component } from 'react';
-import logo from 'assets/logo.png';
+import Routes from 'routes';
+import { ToastContainer } from 'react-toastify';
+import 'react-toastify/dist/ReactToastify.css';
 
 class App extends Component {
   render() {
     return (
       <div className="App">
```
maksymilian-majer: @Jakedavies by doing this you wrapped the whole app in `div.App`, which changed the styling ;)

see on local:
screen shot 2018-07-03 at 15 54 23

vs demo:
screen shot 2018-07-03 at 15 54 31

I'll do a quick-fix on master :)

PR #240, opened 2018-06-30
smart contract setup refactoring and cleanup

### LINKED ISSUE #234: Cleanup blockchain-setup.js in tests
Some cleanup there would do good for future smart contract updates

sindelio:
Closes #234 
This PR adds a file named "contract-bytecode.js" along the "contract-abi.json" to use wherever needed throughout the system. That we we don't have to copy the compiled code of the smart contract in many places, which is error prone. The tests are updated as well


cg/scripts/deploy-blockchain-contract.js
```diff
@@ -1,14 +1,12 @@
...
+const contractABI = require('../src/utils/blockchain/contract-abi.json');
```
maksymilian-majer: I prefer to use camel case even for abbreviations, so like:
`const contractABI` => `const contractAbi`

any preference @Jakedavies?

cg/src/domains/management/contract/contract.service.js
```diff
@@ -17,23 +17,14 @@ async function addressToName(address) {
...
-   */
-  async deployContract(abiJson, compiledData) {
-    const address = await deployContract(abiJson, compiledData);
-
+  async deployContract(contractABIJson, contractByteCode) {
+    const address = await deployContract(contractABIJson, contractByteCode);
     const value = JSON.stringify({
-      abiJson: JSON.stringify(abiJson),
-      compiledData,
+      contractABIJson: JSON.stringify(contractABIJson),
```
maksymilian-majer: same here: `contractABIJson` = > `contractAbiJson `

it feels awkward, and otherwise, it should have been `contractABIJSON` which looks even weirder

but maybe that's just me ;)

cg/src/domains/processors/processors.listeners.js
```diff
@@ -2,15 +2,15 @@ const { listenForErasureRequest, listenForConsent } = require('./../../utils/blo
 const { getDataForSubject } = require('./processors.requests');
 const { blockUntilContractReady } = require('./processors.helpers');
 const { inControllerMode } = require('../../utils/helpers');
-const SubjectService = require('./../subjects/subjects.service');
+const SubjectSService = require('./../subjects/subjects.service');
```
maksymilian-majer: And what's happened with the casing here?
Was anything wrong with `SubjectService`?

cg/scripts/deploy-blockchain-contract.js
sindelio: oks, changing it

cg/src/domains/processors/processors.listeners.js
sindelio: was missing an 's', but I put an uppercase one, fixing it

PR #241, opened 2018-06-30
Fix controller not listed as processor in events and change subject data statuses for more clarity

### LINKED ISSUE #234: Cleanup blockchain-setup.js in tests

sindelio:
Closes #240 Closes #234

notShareable => unconsented
shareable => consented

These statuses are more appropriate, since the controller can't really share the subjects data without their consent.
Controller is now registered as a processor in events that fire with processors as parameters.
Tests updated


sindelio:
Oks

PR #250, opened 2018-07-03
Refactor processor form

### LINKED ISSUE #206: Add new processor
New processors should be added using the same form as the edition. The button design for adding new processors is provided by `kitchen-sink`

fjcero:
Refactored to use the same approach than Users components and increase modularity when adding new actions like delete.

Closes #206


admin/src/containers/Processors/EditProcessor.js
```diff
@@ -0,0 +1,66 @@
...
+      return processor.scopes[s] ? scopes.concat(s) : scopes;
+    }, []);
+
+    return this.props
+      .updateProcessor(processorData)
+      .then(() => this.props.closePanel())
+      .catch(e =>
+        this.setState({
+          errors: {
+            username: e.toString()
```
maksymilian-majer: username?

fjcero: That's true!

PR #252, opened 2018-07-04
Processor delete

### LINKED ISSUE #205: Remove processors
When removing a processor it should confirm the action and should display a feedback on what is happening in the server side in a transparent way to keep users data protected.

fjcero:
Closes #205

### DIFF
--- a/admin/src/components/Processors/Processors.js
+++ b/admin/src/components/Processors/Processors.js
@@ -8,15 +8,20 @@ const Processors = ({
   processors,
   onCreateProcessorClick,
   onEditProcessorClick,
+  onDeleteProcessorClick,
   isLoading,
   children
 }) => {
-  function onCreateProcessorHandler(e) {
+  function onCreateProcessorHandler() {
     onCreateProcessorClick && onCreateProcessorClick();
   }
 
-  function onEditProcessorHandler(e, processor) {
+  function onEditProcessorHandler(processor) {
     onEditProcessorClick && onEditProcessorClick(processor);
+  }
+
+  function onDeleteProcessorHandler(processor) {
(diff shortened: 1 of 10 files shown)


admin/src/components/Processors/Processors.js
```diff
@@ -8,17 +8,22 @@ const Processors = ({
   processors,
   onCreateProcessorClick,
   onEditProcessorClick,
+  onDeleteProcessorClick,
   isLoading,
   children
 }) => {
-  function onCreateProcessorHandler(e) {
+  function onCreateProcessorHandler() {
     onCreateProcessorClick && onCreateProcessorClick();
```
maksymilian-majer: @Jakedavies had a suggestion to make the props required and not check if they exist – I fixed it in a later commit (see master)

admin/src/components/Processors/Processors.js
```diff
@@ -8,17 +8,22 @@ const Processors = ({
...
   }
 
-  function onEditProcessorHandler(e, processor) {
+  function onEditProcessorHandler(processor) {
     onEditProcessorClick && onEditProcessorClick(processor);
```
maksymilian-majer: here too

admin/src/components/Processors/Processors.js
```diff
@@ -8,17 +8,22 @@ const Processors = ({
...
   }
 
+  function onDeleteProcessorHandler(processor) {
+    onDeleteProcessorClick && onDeleteProcessorClick(processor);
```
maksymilian-majer: and here

admin/src/components/Processors/Processors.js
fjcero: 👍

PR #255, opened 2018-07-04
Migrate informed

### LINKED ISSUE #209: Migrate to use informed instead of react-form
This is mentioned in https://github.com/ClearGDPR/ClearGDPR/issues/182#issuecomment-400549848.

- [ ] Normalize structure of all related form components under `Core/Commons/Form`
- [ ] Improve default styling for all controllers and interactions
- [ ] Test interactions in the form
- [ ] Define new components: Checkbox, Textarea

### LINKED ISSUE #279: Not possible to add a new processor in admin UI
Need to provide more info on this, but putting this here for safekeeping.

fjcero:
Closes #209, closes #279


admin/src/components/Processors/EditProcessor.test.js
```diff
@@ -30,8 +37,9 @@ describe('(Component) Edit Processor Form', () => {
     const onSubmit = jest.fn();
     const { component } = setupShallow({ onSubmit });
 
-    const form = component.find('form').at(0);
-    form.simulate('submit');
-    expect(onSubmit).toHaveBeenCalled();
+    const button = component.find('button[type="submit"]');
+    console.log(button.debug());
+    // button.simulate('click');
```
maksymilian-majer: looks unfinished, weren't you able to make it work with informed and the new lack of `<form>`?

admin/src/components/Users/ChangePassword.test.js
```diff
@@ -40,7 +40,7 @@ describe('(Component) Change Password', () => {
 
     const form = component.find('form').at(0);
     form.simulate('submit');
-    expect(onSubmit).toHaveBeenCalled();
+    // expect(onSubmit).toHaveBeenCalled();
```
maksymilian-majer: Same here, if we comment this out, then we don't need the test at all. And maybe we don't since it looks like testing `informed` behavior. Before we were testing our `<form>` to cause `onSubmit` to be called. now I don't think we need this.

admin/src/components/Users/Register.test.js
```diff
@@ -39,8 +41,8 @@ describe('(Component) Change Password', () => {
     const { component } = setupShallow({ onSubmit, validatePassword });
 
     const form = component.find('form').at(0);
-    form.simulate('submit');
-    expect(onSubmit).toHaveBeenCalled();
+    // form.simulate('submit');
```
maksymilian-majer: same here – either we make it work or remove these tests

maksymilian-majer:
That's a big one 🙉 Looking good, but I'll do some manual QA before approving 😉

admin/src/components/Processors/EditProcessor.js
```diff
@@ -1,24 +1,23 @@
...
-    isLoading: PropTypes.bool,
     onSetValues: PropTypes.func,
-    onSubmit: PropTypes.func,
     values: PropTypes.object,
     touched: PropTypes.object,
     errors: PropTypes.object
   };
 
   componentDidMount() {
+    // TODO: This should be probably a helper function, and define a domain entity
```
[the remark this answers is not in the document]
fjcero: 👍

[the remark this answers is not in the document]
fjcero: Created, added to the backlog.

PR #260, opened 2018-07-05
Feature/list subjects widget

### LINKED ISSUE #243: Create widget with table of subjects
- [x] data requesting container
- [x] loading indictator
- [x] pagination
- [x] modify subjects/list endpoint to provide id and created_at, with data in seperate blob
- [x] configurable data render fields?
- [x] created parent container to contain all the "widgets" for the primary dashboard

Jakedavies:
closes #243


admin/src/components/Dashboard/Widgets/SubjectsList.js
```diff
@@ -0,0 +1,33 @@
...
+
+const SubjectsList = ({ subjects, isLoading, errorState }) => {
+  const header = ['id', 'firstname', 'email', 'consented_on'];
+  // TODO: loading?1
+
+  // once we start reaching into data here, this code becomes janky since data is not ours
+  const rows = subjects.map(subject => {
+    return {
+      id: subject.id,
+      consented_on: format(new Date(subject.createdAt), 'DD/MM/YYYY'),
```
maksymilian-majer: hmm.. I used moment in my case, but maybe that's an overkill ;)

PR #271, opened 2018-07-08
Add processor listeners for rectification events and a test for it

### LINKED ISSUE #178: Processors should listen to rectification events from the smart contract
Pull the new data when they are notified of a rectification approval

sindelio:
Closes #178 

It's ready to receive reviews, works on Postman and all tests are passing!

### DIFF
--- a/cg/src/domains/subjects/subjects.service.js
+++ b/cg/src/domains/subjects/subjects.service.js
@@ -13,7 +13,6 @@ const {
 } = require('../../utils/blockchain');
 const { ValidationError, NotFound } = require('../../utils/errors');
 const winston = require('winston');
-
 const { RECTIFICATION_STATUSES } = require('./../../utils/constants');
 const { inControllerMode } = require('./../../utils/helpers');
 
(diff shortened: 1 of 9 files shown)


cg/src/domains/subjects/subjects.service.js
```diff
@@ -29,27 +28,23 @@ class SubjectsService {
   }
 
   async registerConsentToProcessData(subjectId, personalData, processorIds = []) {
+    let processorIdsWithAddresses;
     await this.db.transaction(async trx => {
```
maksymilian-majer: We could refactor this out to a separate method that returns consented processors, but since we're likely to change this approach, then we should do that later.

cg/src/domains/subjects/subjects.service.js
```diff
@@ -103,7 +95,6 @@ class SubjectsService {
   async _createNewSubject(personalData, trx, subjectId) {
     const encryptionKey = generateClientKey();
     const encryptedPersonalData = encryptForStorage(JSON.stringify(personalData), encryptionKey);
-
```
maksymilian-majer: Not sure why you removed all the white-spacing. Just like with writing e-mails or documents, whitespace between paragraphs improves readability. I don't mind this now, but whitespaces don't cost us much, but where it makes sense, I'd like to keep them.
This isn't a necessary change though.

cg/test/processors/start-all.spec.js
```diff
@@ -9,8 +9,8 @@ const processorHelpers = require('../../src/domains/processors/processors.helper
 const { startAll } = require('../../src/domains/processors/processors.listeners');
 
 beforeEach(() => {
-  blockchain.listenForConsent.mockClear();
-  blockchain.listenForErasureRequest.mockClear();
+  blockchain.listenerForConsentEvent.mockClear();
```
maksymilian-majer: we could use `jest.clearAllMocks()`

cg/src/domains/subjects/subjects.service.js
sindelio: I also think we should refactor this service, because it's hard to maintain this piece of the code. Every time I need to work on it, I have to go about understanding it all again.
In the task to create the update consent endpoint, I will try to separate it in 2 methods

cg/src/domains/subjects/subjects.service.js
sindelio: Not all whites-pacing, just the ones between lines of the same blocks of code. There's spaces after db queries, functions, tests and other blocks of code

cg/test/processors/start-all.spec.js
sindelio: Done

PR #281, opened 2018-07-10
Feature/admin/rectification details

### LINKED ISSUE #272: Rectification request details
As a management user I want to see the details of a rectification request after clicking the **Details** button on the list. The details should show past data, the change in data (maybe also data after chage), reason for change and an Approve button.

Optionally there may be a Reject button (if that's supported)

We should either use the panel or a details page => let's see how it looks first.

maksymilian-majer:
Closes #272


fjcero:
I use `import toJson from 'enzyme-to-json'` to reduce the size and improve readability of snapshots. If you will that is a good approach when updating the missing snapshot :)

maksymilian-majer:
@fjcero I rewrote to use `enzyme-to-json` and I think in most cases it's probably better to use it as default serializer. I added an issue #283 to consider if that's a good idea – there are also non-component snapshots after all.

PR #287, opened 2018-07-11
Findsecrets update

maksymilian-majer:
Updated findsecrets and ignored false-positive lines.
It's related to https://github.com/clevertech/DevOps/issues/10035

### DIFF
--- a/cg/src/utils/encryption.js
+++ b/cg/src/utils/encryption.js
@@ -11,7 +11,7 @@ function mergeTypedArray(a, b) {
 }
 
 function encryptWithNonce(message, key) {
-  const nonce = sodium.randombytes_buf(sodium.crypto_secretbox_NONCEBYTES);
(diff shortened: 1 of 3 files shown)


cg/src/utils/encryption.js
```diff
@@ -11,7 +11,7 @@ function mergeTypedArray(a, b) {
 }
 
 function encryptWithNonce(message, key) {
-  const nonce = sodium.randombytes_buf(sodium.crypto_secretbox_NONCEBYTES);
+  const nonce = sodium.randombytes_buf(sodium.crypto_secretbox_NONCEBYTES); // finsecrets-ignore-line
```
Jakedavies: finsecrets? is that correct?

maksymilian-majer:
@Jakedavies not my typo, but a typo nonetheless
I'll report it here https://github.com/clevertech/findsecrets

gimenete:
Fixed in version `0.1.3`. Thanks for reporting it :)

PR #301, opened 2018-07-13
Fix/admin/this bindings

### LINKED ISSUE #285: Refactor handlers to bind this more appropriately
We want to use this approach:

```
handleChange = () => {
  // call this function from render 
  // and this.whatever in here works fine.
};
```

More info is here: https://medium.freecodecamp.org/react-binding-patterns-5-approaches-for-handling-this-92c651b5af56

maksymilian-majer:
Closes #285


Jakedavies:
I think I broke master... checking why, it seems to be passing locally

Jakedavies:
I just made a commit directly to master, should fix the issue.

Jakedavies:
Its passing now, just hit the caching issue it looks like.

PR #313, opened 2018-07-16
removes "integration" test runner

### LINKED ISSUE #45: Remove the test:integration differentiator and documentation
All our tests are integration so we should remove the option to split those tests, cause it's not used.

Jakedavies:
closes #45


Jakedavies:
closes

Jakedavies:
removed all the appUnit calls and stuff too. Everything looks good now.

PR #324, opened 2018-07-17
Share data (Right to data portability)

### LINKED ISSUE #242: Create HTML for share URL interface

fjcero:
Close #242 

image


maksymilian-majer:
@fjcero I can see an `undefined` there on the screenshot ;)

cg/src/domains/subjects/data-shares.service.js
```diff
@@ -8,8 +8,10 @@ class DataShareService {
     return db('data_shares')
       .where({ subject_id: subjectId })
       .map(dataShare => ({
+        id: dataShare.id,
         name: dataShare.name,
-        url: `${process.env.url}/api/subjects/data-shares/share?token=${dataShare.token}`
+        token: dataShare.token,
+        url: `${process.env.ROOT_URL}/api/subjects/data-shares/share?token=${dataShare.token}`
```
maksymilian-majer: I think we should use the `CONTROLLER_URL` here instead or we need to update the quick start scripts.

fjcero:
@maksymilian-majer yup! that is the thing I found in the CG API. It seems the URL is build using an env var. I defined a new `ROOT_URL` variable, but that should be enforced in the `env.example` and `node setup.js` as well. So first probably check if that variable name is ok or we should follow another approach.

frontend/src/js-sdk/modules/subject.js
```diff
@@ -24,4 +29,22 @@ export default class Subject extends ResourceBase {
   getProcessors() {
     return this.request(`${SUBJECT}/processors`);
   }
+
+  getSharedData(token) {
+    return this.request(`${SUBJECT}/data-shares/share?token=${token}`);
+  }
+
+  getDataShares(token) {
```
maksymilian-majer: `token` is not used

frontend/src/elements/components/ShareData.js
```diff
@@ -0,0 +1,282 @@
...
+  </React.Fragment>
+);
+
+class SharedData extends React.Component {
+  state = {
+    shared: {}
+  };
+
+  componentDidMount() {
+    window.cg.Subject.getSharedData(this.props.sharedDataToken)
```
maksymilian-majer: are we actually using this endpoint in the UI?
this is for 3rd parties to use IMO

fjcero: Ouh, I see. I supposed that, but either way, created an specific component to show which data is shared like this:

image

frontend/src/js-sdk/modules/subject.js
fjcero: 👍

cg/src/domains/subjects/data-shares.service.js
fjcero: The config generated by `node setup.js` has the value `CONTROLLER_URL=http://cg:8082`. Is generating in different way on production deploys?

frontend/src/elements/components/ShareData.js
```diff
@@ -0,0 +1,282 @@
...
+              </span>
+            </a>
+            <div>
+              <span className={styles.shareImg}>
+                <img src={clipboardSvg} alt={share.name} style={{ height: 35 }} />
+              </span>
+              <div className={styles.shareName}>{share.name}</div>
+              <div className={styles.shareUrl}>{share.url}</div>
+              <SharedData sharedDataToken={share.token} />
+              {/* <button
```
fjcero: Edit doesn't make sense, but I added that since it was in the design first, and then check that perhaps don't make to much sense, at least Rectification has something to do with it.

frontend/src/elements/components/ShareData.js
maksymilian-majer: @sindelio @Jakedavies can you chime in here? I think we're counting how many times the URL has been used, are we not? In that case I'd rather have this removed from here and we can get the data from a different endpoint: 
`/api/subject/access-data`

frontend/src/elements/components/ShareData.js
```diff
@@ -59,7 +59,7 @@ const Delete = ({ onDelete, onCancel }) => (
       </button>
       &nbsp;
       <button onClick={onDelete} className="button is-danger">
-        I'm sure
+        {"I'm sure"}
```
maksymilian-majer: makes the PR read so much nicer ;)

cg/src/domains/subjects/data-shares.service.js
maksymilian-majer: No, there's no automation yet for production deployments, but our demo API has manually generated proper env vars. If you change it to `ROOT_URL` we'll need to include it in `setup.js` and ask DevOps for update on Kubernetes.

frontend/src/elements/components/ShareData.js
fjcero: Soooo true

cg/src/domains/subjects/data-shares.service.js
fjcero: Let's do it `CONTROLLER_URL` then! 👍

frontend/src/elements/components/ShareData.js
fjcero: I will remove this for now. In next iterations we can decide how is better to allow subjects to share information, since for example they should be able to rectify and also allow/disallow which data they are sharing with kind of "settings" for each share.

maksymilian-majer: Have you removed it after all?

PR #345, opened 2018-07-19
Fix/processors domain reaching into subjects domain

### LINKED ISSUE #311: Call recordAccessByController in the access-data endpoint
That endpoint is missing the function that acts inside the smart contract.
It turns out it's not going to be so simple because the processors code is reaching into the subjects domain.
Blocked by #329

### LINKED ISSUE #329: Fix processors domain reaching into the subjects domain
We have to separate those asap, it only yields problems

sindelio:
Closes #329 Closes #311 

Creates a processors service in the processors domain, separating the processors domain from the subjects domain. They have essentially the same code with a few slightly differences.
Also calls the smart contract from the data-access endpoint in the subjects domain, which was missing.
Tests are passing, but I want to test it more in Postman before considering merging.
Anyways it's good to start receiving reviews

### DIFF
--- a/cg/src/domains/processors/processors.controller.js
+++ b/cg/src/domains/processors/processors.controller.js
@@ -10,7 +10,7 @@ class ProcessorsController {
 
   async getSubjectData(req, res) {
     const subjectId = req.params.subjectId;
-    const subjectData = await this.processorsService.getData(subjectId);
+    const subjectData = await this.processorsService.getSubjectData(subjectId);
     return res.json(subjectData);
   }
 
--- a/cg/src/domains/processors/processors.controller.js
+++ b/cg/src/domains/processors/processors.controller.js
@@ -1,16 +1,16 @@
-const ProcessorsService = require('./processors.service');
+const SubjectsService = require('./processors.service');
 const ContractService = require('./contract.service');
 const { NotFound } = require('../../utils/errors');
 
 class ProcessorsController {
-  constructor(processorsService = null, contractService = null) {
-    this.processorsService = processorsService || new ProcessorsService();
+  constructor(subjectsService = null, contractService = null) {
+    this.subjectsService = subjectsService || new SubjectsService();
     this.contractService = contractService || new ContractService();
   }
 
   async getSubjectData(req, res) {
     const subjectId = req.params.subjectId;
-    const subjectData = await this.processorsService.getSubjectData(subjectId);
+    const subjectData = await this.subjectsService.getSubjectData(subjectId);
     return res.json(subjectData);
   }
 
--- a/cg/src/domains/processors/processors.controller.js
+++ b/cg/src/domains/processors/processors.controller.js
@@ -1,18 +1,16 @@
-const SubjectsService = require('./../subjects/subjects.service');
-const ContractService = require('./../management/contract/contract.service');
+const ProcessorsService = require('./processors.service');
+const ContractService = require('./contract.service');
 const { NotFound } = require('../../utils/errors');
 
-// TODO: maybe we don't want to reach into the subjects / management domain?
-
 class ProcessorsController {
-  constructor(subjectsService = null, contractService = null) {
-    this.subjectsService = subjectsService || new SubjectsService();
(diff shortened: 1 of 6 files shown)


cg/src/domains/processors/processors.controller.js
```diff
@@ -1,18 +1,16 @@
...
+  constructor(processorsService = null, contractService = null) {
+    this.processorsService = processorsService || new ProcessorsService();
     this.contractService = contractService || new ContractService();
   }
 
-  async getData(req, res) {
+  async getSubjectData(req, res) {
     const subjectId = req.params.subjectId;
-    const subjectData = await this.subjectsService.getData(subjectId);
+    const subjectData = await this.processorsService.getSubjectData(subjectId);
```
maksymilian-majer: I'd rather have a `SubjectsService` in processors domain if that method is in the API. The `ProcessorsService` should rather operate on processors and not subjects.

cg/src/domains/processors/processors.controller.js
```diff
@@ -1,18 +1,16 @@
-const SubjectsService = require('./../subjects/subjects.service');
-const ContractService = require('./../management/contract/contract.service');
+const ProcessorsService = require('./processors.service');
```
maksymilian-majer: `const SubjectsService = require('./../subjects/subjects.service');` => `const SubjectsService = require('./subjects.service');`

see comments below

cg/src/domains/processors/processors.listeners.js
```diff
@@ -6,18 +6,18 @@ const {
 const { getDataForSubject } = require('./processors.requests');
 const { blockUntilContractReady } = require('./processors.helpers');
 const { inControllerMode } = require('../../utils/helpers');
-const SubjectsService = require('./../subjects/subjects.service');
+const ProcessorsService = require('./processors.service');
```
maksymilian-majer: same as above

cg/src/domains/processors/processors.service.js
```diff
@@ -0,0 +1,115 @@
+const { db } = require('../../db');
```
maksymilian-majer: Rename to `SubjectsService`, but keep in the same domain.

maksymilian-majer: I think there should be some discussion in place, cause this isn't DRY if we need to maintain two very similar classes in two separate domains. Let's think on how the processors' domain should interact with the subject's data. Maybe we can simplify things.

cg/src/domains/subjects/subjects.service.js
```diff
@@ -104,30 +100,6 @@ class SubjectsService {
     await this._saveSubjectEncryptionKey(trx, subjectId, encryptionKey);
   }
 
-  // This code is being used by the listeners in the processors domain. We should refactor this later to not have inter domain interaction
```
maksymilian-majer: I like that the code is going away. Still I'm considering how processors should act and what should they store to avoid duplication.

cg/src/domains/processors/processors.service.js
sindelio: Ok

cg/src/domains/processors/processors.controller.js
sindelio: I see, it makes more sense. Changing it

cg/src/domains/processors/processors.listeners.js
sindelio: Oks

cg/src/domains/processors/processors.controller.js
sindelio: Oks

cg/src/domains/subjects/subjects.service.js
sindelio: Hmm I can't see a better way to do this. Maybe we could create a subjects service in a domain that's specific to those 2? Like a shared domain between them, with only 1 subjects service.

sindelio: I will bring it up in the next meeting, Monday

PR #349, opened 2018-07-23
Feature/processors listeners for restriction events

### LINKED ISSUE #346: Create processors listener for restriction events
Processors should be able to listen to restriction events fired by the controller and update their subjects data accordingly

sindelio:
Closes #346 

Creates the processors listener for restriction events.
It's tested in Postman, and the there's a working test for it too.
It's good to receive reviews o/
Note: Took me more than 5 hours fighting with a bug to find and fix it, but it all worked out :))

### DIFF
--- a/cg/src/domains/processors/processors.requests.js
+++ b/cg/src/domains/processors/processors.requests.js
@@ -14,6 +14,15 @@ async function getDataForSubject(subjectId) {
   });
 }
 
+async function getRestrictionsForSubject(subjectId) {
+  return await fetch(`${CONTROLLER_URL}/api/processors/subject/${subjectId}/get-restrictions`, {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${PROCESSOR_JWT}`
+    }
+  });
+}
+
 async function getContractDetails() {
   try {
     const res = await fetch(`${CONTROLLER_URL}/api/processors/contract/details`, {
@@ -42,5 +51,6 @@ async function getContractDetails() {
 
 module.exports = {
   getDataForSubject,
+  getRestrictionsForSubject,
   getContractDetails
 };
--- a/cg/src/domains/processors/processors.routes.js
+++ b/cg/src/domains/processors/processors.routes.js
@@ -13,6 +13,11 @@ module.exports = app => {
   router.use(verifyJWT);
 
   router.get(
+    '/contract/details',
+    asyncHandler(async (req, res) => processorsController.getContractDetails(req, res))
+  );
+
+  router.get(
     '/subject/:subjectId/data',
     controllerOnly,
     asyncHandler(ensureProcessorAccessToSubject),
@@ -20,7 +25,9 @@ module.exports = app => {
   );
 
   router.get(
-    '/contract/details',
-    asyncHandler(async (req, res) => processorsController.getContractDetails(req, res))
+    '/subject/:subjectId/get-restrictions',
+    controllerOnly,
+    asyncHandler(ensureProcessorAccessToSubject),
+    asyncHandler(async (req, res) => processorsController.getSubjectRestrictions(req, res))
   );
 };
--- a/cg/src/utils/blockchain.js
+++ b/cg/src/utils/blockchain.js
@@ -87,8 +87,18 @@ async function setSubjectDataState(subjectId, processor, state) {
   return await runContractMethod('setSubjectDataState', [subjectId, processor, state]);
 }
 
-async function setSubjectRestrictions(subjectId, directMarketing = false, emailCommunication = false, research = false) {
-  return await runContractMethod('setSubjectRestrictions', [subjectId, directMarketing, emailCommunication, research]);
+async function setSubjectRestrictions(
+  subjectId,
+  directMarketing = false,
+  emailCommunication = false,
+  research = false
+) {
+  return await runContractMethod('setSubjectRestrictions', [
+    subjectId,
+    directMarketing,
+    emailCommunication,
+    research
+  ]);
 }
 
 async function setProcessors(newProcessors = []) {
@@ -123,8 +133,18 @@ async function recordRectificationByController(subjectId) {
   return await runContractMethod('recordRectificationByController', [subjectId]);
 }
 
-async function recordRestrictionByController(subjectId, directMarketing = false, emailCommunication = false, research = false) {
-  return await runContractMethod('recordRestrictionByController', [subjectId, directMarketing, emailCommunication, research]);
+async function recordRestrictionByController(
+  subjectId,
+  directMarketing = false,
+  emailCommunication = false,
+  research = false
+) {
+  return await runContractMethod('recordRestrictionByController', [
+    subjectId,
+    directMarketing,
+    emailCommunication,
+    research
+  ]);
 }
 
 async function recordErasureByController(subjectId) {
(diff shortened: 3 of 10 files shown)


cg/src/domains/processors/processors.requests.js
```diff
@@ -14,6 +14,15 @@ async function getDataForSubject(subjectId) {
   });
 }
 
+async function getRestrictionsForSubject(subjectId) {
+  return await fetch(`${CONTROLLER_URL}/api/processors/subject/${subjectId}/get-restrictions`, {
+    method: 'GET',
```
Nelrohd: it's not GET by default?

cg/src/domains/processors/processors.routes.js
```diff
@@ -20,7 +25,9 @@ module.exports = app => {
   );
 
   router.get(
-    '/contract/details',
-    asyncHandler(async (req, res) => processorsController.getContractDetails(req, res))
+    '/subject/:subjectId/get-restrictions',
+    controllerOnly,
+    asyncHandler(ensureProcessorAccessToSubject),
+    asyncHandler(async (req, res) => processorsController.getSubjectRestrictions(req, res))
```
Nelrohd: I think we can make it shorter here
```
asyncHandler(processorsController.getSubjectRestrictions)
```

cg/test/processors/processors.api.spec.js
```diff
@@ -1,7 +1,7 @@
 jest.mock('../../src/utils/blockchain/web3-provider-factory');
 jest.mock('../../src/utils/helpers');
-const encryption = require('./../../src/utils/encryption');
 
+const encryption = require('./../../src/utils/encryption');
```
Nelrohd: `./../../src/utils/encryption'` => `../../src/utils/encryption'`?

cg/test/processors/processors.listeners.spec.js
```diff
@@ -162,4 +164,40 @@ describe('Processors listening for blockchain events', () => {
...
+      subject_id: subjectId,
+      key: encryptionKey
+    });
+
+    // When
+    await recordRestrictionByController(subjectId, false, true, false);
+
+    // Then
+    setTimeout(async () => {
+      // We have to wait for the processor to update his database with the new subject restrictions
```
Nelrohd: is there a way to force a refresh?

cg/src/utils/blockchain.js
```diff
@@ -176,6 +196,25 @@ async function listenerForRectificationEvent(callback) {
...
+  return await quorumContract.contract.events.Controller_SubjectDataRestricted((error, data) => {
+    if (error) {
+      winston.error(`Error handling restriction ${error.toString()}`);
+      return;
+    }
+
+    // There's a bug when a smart contract event is fired with a boolean parameter set to 'false', which transforms that boolean in 'null'.
+    // This piece of code corrects the nulled booleans, if there is any.
+    // I'm not sure why this bug happens, but it seems to be an issue with event firing in Ethereum/Quorum
+    const directMarketing = data.returnValues.directMarketing === null ? false : true;
```
Nelrohd: `data.returnValues.directMarketing === null ? false : true` => `data.returnValues.directMarketing !== null` ?

cg/src/utils/blockchain.js
```diff
@@ -176,6 +196,25 @@ async function listenerForRectificationEvent(callback) {
   });
 }
 
+async function listenerForRestrictionEvent(callback) {
+  const quorumContract = await getContract();
+  return await quorumContract.contract.events.Controller_SubjectDataRestricted((error, data) => {
```
Nelrohd: you can remove the `await` since you return the promise, because of `async`, the value is always a promise.

cg/src/utils/blockchain.js
```diff
@@ -123,8 +133,18 @@ async function recordRectificationByController(subjectId) {
...
 
-async function recordRestrictionByController(subjectId, directMarketing = false, emailCommunication = false, research = false) {
-  return await runContractMethod('recordRestrictionByController', [subjectId, directMarketing, emailCommunication, research]);
+async function recordRestrictionByController(
+  subjectId,
+  directMarketing = true,
+  emailCommunication = true,
+  research = true
+) {
+  return await runContractMethod('recordRestrictionByController', [
```
Nelrohd: same comment as below (you can remove the `await`)

cg/src/utils/blockchain.js
```diff
@@ -87,8 +87,18 @@ async function setSubjectDataState(subjectId, processor, state) {
...
 
-async function setSubjectRestrictions(subjectId, directMarketing = false, emailCommunication = false, research = false) {
-  return await runContractMethod('setSubjectRestrictions', [subjectId, directMarketing, emailCommunication, research]);
+async function setSubjectRestrictions(
+  subjectId,
+  directMarketing = false,
+  emailCommunication = false,
+  research = false
+) {
+  return await runContractMethod('setSubjectRestrictions', [
```
Nelrohd: same comment as below (you can remove the await)

cg/src/domains/processors/processors.requests.js
sindelio: It is but clarity is always good. We're building a back-end framework for other devs afterall

cg/src/domains/processors/processors.routes.js
sindelio: Gonna take a look at it

cg/src/utils/blockchain.js
sindelio: Trying it out

cg/test/processors/processors.listeners.spec.js
sindelio: It's listening to an event, I can't see another way to refresh the restrictions. There might be, but I can't see

cg/test/processors/processors.api.spec.js
sindelio: Oks

cg/src/domains/processors/processors.routes.js
sindelio: We actually can't, the method getSubjectRestrictions needs to know it's input parameters. 
We can use this shorthand when the middleware does not need any parameter

cg/src/utils/blockchain.js
sindelio: We can't do that. All of those awaits are necessary.
Let me try to explain in simple terms, think of any blockchain call as a database query, it takes time but we need to wait until it finishes. In this case the listener won't be attached if we don't wait till it gets "set" in blockchain land

cg/src/utils/blockchain.js
sindelio: This is a good short-hand, thx

cg/src/utils/blockchain.js
Nelrohd: Ok thanks. I will check if we have a ticket for the refactor.

cg/test/processors/processors.listeners.spec.js
Nelrohd: ok thank you, it was just for my own information

cg/src/domains/processors/processors.requests.js
Nelrohd: I see but the GET as default is fairly common and we should have a API documentation somewhere to make it even clearer.

PR #354, opened 2018-07-24
Sign up form - show error received from backend call

### LINKED ISSUE #332: Show form backend errors in Sign Up form
This happens when signing up with an email that is already registered

segoddnja:
Closes #332

### DIFF
--- a/frontend/src/containers/SignUp.js
+++ b/frontend/src/containers/SignUp.js
@@ -28,13 +28,19 @@ class SignUp extends Component {
       })
     })
       .then(res => res.json())
-      .then(token => {
+      .then(response => {
         this.setState({ isLoading: false });
+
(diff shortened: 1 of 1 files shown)


frontend/src/containers/SignUp.js
```diff
@@ -73,6 +80,7 @@ class SignUp extends Component {
               <p>{`To create your Account simply fill the short form below`}</p>
               <hr />
               <form onSubmit={e => this.onSignUpHandler(e)}>
+                {error ? <div className="notification is-danger">{error}</div> : null}
```
fjcero: It could be rewritten to conditional rendering form `error && <div />`. Also, having a default value for error state makes the code easier to read, so adding that to the `state = { error: null }` will be helpful

PR #357, opened 2018-07-26
Complete right to objection

### LINKED ISSUE #356: Create the right to objection endpoints, processor listener and test
- [x] Create smart contract support for objections
- [x] Create 2 endpoints for subjects

/api/subject/object
/api/subject/get-objection

- [x] Add a migration in the database subjects table
- [x] Create the processors listener for objection events
- [x] Add tests for the endpoints and the listener

sindelio:
Closes #356 

Creates:

- Smart contract support for the right to objection
- Endpoints that allows subjects to object to the processing of their data
    /api/subject/object
    /api/subject/get-objection

- A migration to the 'subjects' table for the objection status
- The processors listener for objection events
- Tests for the endpoints and listener

Ready to receive reviews o/


cg/src/domains/subjects/subjects.controller.js
```diff
@@ -16,10 +16,7 @@ class SubjectsController {
   }
 
   async updateConsent(req, res) {
-    await this.subjectsService.updateConsent(
-      req.subject.id,
-      req.body.processors || []
-    );
+    await this.subjectsService.updateConsent(req.subject.id, req.body.processors || []);
     res.json({ success: true });
```
Nelrohd: is this a standard to send `success` or could we just use HTTP code 204 to notify there is nothing to send but the request was successful?

PR #358, opened 2018-07-26
Feature/request to be forgotten

### LINKED ISSUE #176: Improve "request to be forgotten" flow
**Description:**

As a subject when I go to the demo front-end and want to request to erase my data I have to click a button.

After subject clicks the button the system needs to show them a warning message, that when they erase thair data all the share URLs will become inactive and they will no longer be able to retrieve their data from our system.

**Acceptance criteria**

- [ ] Add confirmation when clicking the "request" button
- [ ] Add a test to cover this use case
- [ ] Shared data should be unavailable after accepting deletion

**Steps:**

1. In `demo-frontend` app
2. I login as a registered user
3. Click the "request to be forgotten" button
4. There should be a confirmation message shown with "Are you sure you want to request erasure of your data?"
5. If the user goes to the `/shares` page should appear all shared data as inactive (or deleted)

segoddnja:
Implemented next features for demo frontend:
- global subject context to get access to data and status
- high order component to wrap specific components and display `data erased` screen when it is necessary
- boilerplate `loading` and `data erased` screens

closes #176


segoddnja:
@fjcero I was not sure about proper location for `SubjectContext` and `DataErasureAwareContainer` so if you can advice better place - I'm all ears :)

frontend/src/Elements/components/Common/Views/DataErasedScreen.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+export default () => (
+  <React.Fragment>
```
Nelrohd: I'm not sure we need React Fragment since section is the only child

frontend/src/Elements/components/Common/Views/LoadingScreen.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ModalView from './Modal';
+import styles from '../../../theme/ForgottenRequest.scss';
+
+export default () => (
+  <React.Fragment>
```
Nelrohd: Same comment for React.Fragment

frontend/src/containers/SubjectContext.js
```diff
@@ -0,0 +1,147 @@
...
+  isErased: false,
+  data: null,
+  status: null
+};
+
+let originalLocalStorageSetItem;
+const EVENT_NAME = 'localStorage.itemInserted';
+const LOCAL_STORAGE_EVENTS = [EVENT_NAME, 'storage'];
+
+function upgradeLocalStorage() {
```
Nelrohd: I think it's better to stick to ```const functionName () => {``` to stay consistent with the above code

frontend/src/containers/SubjectContext.js
```diff
@@ -0,0 +1,147 @@
...
+  if (originalLocalStorageSetItem) {
+    return;
+  }
+
+  originalLocalStorageSetItem = localStorage.setItem;
+  localStorage.setItem = function() {
```
Nelrohd: ES6 -> ```() => {```

frontend/src/containers/SubjectContext.js
```diff
@@ -0,0 +1,147 @@
...
+    originalLocalStorageSetItem.apply(this, arguments);
+
+    const keyboardEvent = new CustomEvent(EVENT_NAME);
+    keyboardEvent.key = arguments[0];
+    keyboardEvent.value = arguments[1];
+    window.dispatchEvent(keyboardEvent);
+  };
+}
+
+function downgradeLocalStorage() {
```
Nelrohd: arrow function also

frontend/src/containers/SubjectContext.js
```diff
@@ -0,0 +1,147 @@
...
+export class SubjectProvider extends React.Component {
+  static propTypes = {
+    children: PropTypes.oneOfType([
+      PropTypes.element,
+      PropTypes.object,
+      PropTypes.arrayOf(PropTypes.element)
+    ])
+  };
+
+  _fetchData = async () => {
```
Nelrohd: what is the norm for using `_` as prefix?

frontend/src/containers/SubjectContext.js
```diff
@@ -0,0 +1,147 @@
...
+    let data;
+    let status;
+
+    try {
+      status = await window.cg.Subject.getDataStatus();
+    } catch (e) {
+      status = null;
+    }
+
+    const isErased = !status || status.controller === 2;
```
Nelrohd: We could use constants instead of "2" (DELETED)

segoddnja:
@fjcero I think it makes sense to move `SubjectContext` to Elements SDK scope and interact with `CG` backend only trough context. Another question is what to do with JS SDK. In case we decide to invest all time to develop Elements SDK first, does it make sense to make it dependent on JS SDK or not? My opinion is that we can build JS SDK for internal usage with possibly unstable interface, and with only methods we need for Elements SDK to work. Later, when we decide to release JS SDK for clients we will be able to do it easily without need to keep backward compatibility and without need to do a lot of changes in Elements SDK.

PR #380, opened 2018-08-08
Add new element to Element SDK - Objection

### LINKED ISSUE #368: Right to objection of processing
Allow the Subject to object processing to the controller or certain processor

Ref. Article: https://gdpr-info.eu/art-21-gdpr/

Create a new Element SDK component that renders a way to allow the Subject to prevent processing of data. For now just closing the loop will be enough, no need to address to much granularity control.

The endpoints used for Objection are:
- `/api/subject/get-objection`
- `/api/subject/object`

**Additional context**
No design provided, so first, we should address interaction and then some styling

segoddnja:
Add this element to demo page
Modify Switch element - add `disabled` state ssupport
Add new methods to JS SDK to support interaction with Objection

[Screenshot](http://pix.toile-libre.org/upload/original/1533726040.png)

closes #368

### DIFF
--- a/frontend/src/Elements/components/Objection/Objection.js
+++ b/frontend/src/Elements/components/Objection/Objection.js
@@ -1,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Switch from '../Common/Switch';
+
+class Objection extends React.PureComponent {
+  state = {
+    allowDataProcessing: true,
+    busy: true
+  };
+  async componentDidMount() {
+    const { objection } = await this.props.cg.Subject.getObjectionStatus();
+    this.setState({ busy: false, allowDataProcessing: !objection });
+  }
+  async toggleDataProcessingAgreement() {
+    const { busy, allowDataProcessing } = this.state;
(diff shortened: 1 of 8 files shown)


frontend/src/Elements/components/Objection/Objection.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  render() {
+    const { label = 'Object data processing' } = this.props.options;
+    const { allowDataProcessing, busy } = this.state;
+
+    return (
+      <div style={{ alignItems: 'center', display: 'flex' }}>
+        <label>{label}</label>&nbsp;
+        <Switch
+          onChange={this.toggleDataProcessingAgreement.bind(this)}
```
fjcero: Just noticed this, if you want to avoid the `bind(this)` stuff, you can declare the function as a class property, that with the help of Babel it gets compiled automatically. We're using that syntax in the `admin` but can be done here as well.

segoddnja: Thanks, I'll check.

PR #396, opened 2018-08-15
Create endpoint to fetch blockchain events

### LINKED ISSUE #388: Endpoint for fetching blockchain events + basic filtering
We can't fetch previous blockchain events from the API.

Have an endpoint to fetch previous blockchain events and basic filtering.

None.

**Additional context**
None.

segoddnja:
closes #388


cg/src/domains/management/stats/stats.service.js
```diff
@@ -51,6 +52,10 @@ class StatsService {
       processors: processorData
     };
   }
+
+  async events(eventType = 'allEvents', fromBlock = 0, toBlock = 'latest', filter) {
+    return await blockchain.getPastEvents(eventType, fromBlock, toBlock, filter);
```
Nelrohd: if blockchain.getPastEvents return a promise, we don;t need to use async/await in this case

cg/src/utils/blockchain.js
```diff
@@ -174,9 +174,9 @@ async function recordErasureByProcessor(subjectId) {
 
 // AUDITING FUNCTION
 
-async function getPastEvents(event) {
+async function getPastEvents(eventType, fromBlock, toBlock, filter) {
   const quorumContract = await getContract();
-  return await quorumContract.contract.getPastEvents(event, { fromBlock: 0, toBlock: 'latest' });
+  return await quorumContract.contract.getPastEvents(eventType, { fromBlock, toBlock, filter });
```
Nelrohd: same as above

cg/smart_contracts/ClearGDPR.sol
```diff
@@ -31,13 +31,13 @@ contract ClearGDPR {
...
-    event Controller_SubjectDataAccessed(bytes32 subjectId);
-    event Controller_SubjectDataRectified(bytes32 subjectId, uint256 rectificationCount);
-    event Controller_SubjectDataRestricted(bytes32 subjectId, bool directMarketing, bool emailCommunication, bool research);
-    event Controller_SubjectDataObjected(bytes32 subjectId, bool objection);
-    event Controller_SubjectDataErased(bytes32 subjectId); 
-    event Processor_SubjectDataErased(bytes32 subjectId, address processor);
+    event Controller_ConsentGivenTo(bytes32 indexed subjectId, address[] processorsConsented);
+    event Controller_SubjectDataAccessed(bytes32 indexed subjectId);
+    event Controller_SubjectDataRectified(bytes32 indexed subjectId, uint256 rectificationCount);
+    event Controller_SubjectDataRestricted(bytes32 indexed subjectId, bool directMarketing, bool emailCommunication, bool research);
```
sindelio: Did you test these changes in the smart contract? 
Solidity code is very delicate!

segoddnja: Yes. After my first change some tests failed(because I tried to make address[] indexed). So I had to revert it back.

cg/src/domains/management/management.routes.js
```diff
@@ -99,6 +99,8 @@ module.exports = app => {
     asyncHandler(async (req, res) => subjectsController.getRectificationRequest(req, res))
   );
 
+  router.get('/events', asyncHandler(async (req, res) => statsController.events(req, res)));
```
sindelio: Change to /past-events, that way it's clear we are talking about the past events and not present events

cg/src/utils/blockchain/contract-bytecode.js
```diff
@@ -1,4 +1,4 @@
 const contractByteCode =
```
sindelio: Where did you get this compiled code?
When u change the bytecode, you also have to change the ABI (Application Binary Interface)

cg/test/management/stats.api.spec.js
```diff
@@ -135,4 +154,73 @@ describe('Stats endpoint', () => {
       })
     );
   });
+
+  it('should return events', async () => {
```
sindelio: should return the past events

cg/test/management/stats.api.spec.js
```diff
@@ -41,6 +41,25 @@ beforeEach(() => {
...
+  await clearEvents();
+  let r;
+  for (let i in SUBJECTS) {
+    r = await recordConsentGivenTo(SUBJECTS[i], []);
+  }
+  return r.blockNumber;
+}
+
 describe('Stats endpoint', () => {
   it('Should display 0 stats properly if there is no entries yet', async () => {
```
sindelio: there are no entries, fix this too please

cg/test/management/stats.api.spec.js
```diff
@@ -135,4 +154,73 @@ describe('Stats endpoint', () => {
...
+      }
+    });
+    const events = await res.json();
+    expect(events).toHaveLength(1);
+    expect(events[0].blockNumber).toBe(lastBlockNumber);
+  });
+
+  it('should allow to get events before some certain event', async () => {
+    const lastBlockNumber = await prepareEvents();
+    const firstBlockNumber = lastBlockNumber - SUBJECTS.length + 1;
```
sindelio: Won't firstBlock always be 0?

cg/src/domains/management/stats/stats.service.js
sindelio: I like to use it tho, makes the code clearer about what is a promise and what isn't

sindelio: For instance, we also don't need to return in the last line of a function, but we do so for clarity

cg/src/domains/management/management.routes.js
segoddnja: I do not think that it makes to concentrate attention that events are "past". This is http endpoint so it by definition can return only "past" events, so this word in the endpoint name is redundant.

cg/src/utils/blockchain/contract-bytecode.js
segoddnja: I used https://remix.ethereum.org to compile contract. Actually, we need to make compilation to be a part of contract deploy process. I am not sure that keep compiled ABI and bytecode in the repo is a good idea either.

cg/test/management/stats.api.spec.js
segoddnja: See my comment above

cg/test/management/stats.api.spec.js
segoddnja: I did not review existed tests. Ok, will fix this, no problem.

segoddnja:
@sindelio I think it makes sense to create test suite which directly covers smart contract functionality. I know that right now we have a bunch of e2e tests that transitively covers smart contract, but, it's not the same.

cg/test/management/stats.api.spec.js
sindelio: Yeah that's right

sindelio:
Ideally yes, but we are including the smart contract function calls into the e2e tests, so we should be good for now. We are also testing everything manually before committing and merging. If we get more time ahead of September 15, then we can work on this, right now we have things with a higher priority

cg/test/management/stats.api.spec.js
sindelio: Thx

cg/src/utils/blockchain/contract-bytecode.js
sindelio: Alright, we're good then

cg/src/domains/management/management.routes.js
sindelio: An uniformed reader would not know that. I vote for clarity

cg/smart_contracts/ClearGDPR.sol
sindelio: There's a bunch of traps in Solidity, gotta watch out and test every move!

cg/src/domains/management/management.routes.js
segoddnja: Would not know what? The fact that HTTP endpoint provides access to collection of some entities? Why do not we name all endpoints in the same way - `/api/management/created-processors`, `/api/management/existed-subjects` or something like this?

Btw, about endpoint names. Are we trying to build restful api or not? At this moment it looks more like not is an answer. An example

```
GET /api/management/processors/list
POST /api/management/processors/add
POST /api/management/processors/update
POST /api/management/processors/delete
```

The proper implementation is to use the same endpoint `/api/management/processors` with GET, POST, PUT and DELETE methods. Parts `list`, `add`, `update` and `delete` do not provide any additional clarity.

sindelio: Ok you have a point about the events. 
About the API, it wasn't me who made the decision to make it like this. But in my opinion it's better to have the actual action in the endpoint, makes it easier to read. Not everything fits in GET, POST, PUT and DELETE. Anyways, would be unnecessary work to change it now

sindelio:
All comments were addressed, it's good to merge

cg/src/domains/management/management.routes.js
segoddnja: I do not call to change it right now, I just say what is the proper way. I agree that not everything fits in CRUD operations, but, example is 100% CRUD and operation names in endpoints are redundant. Just to consider it as something that can be improved later.

PR #401, opened 2018-08-16
Feature/data attributes configuration endpoint

### LINKED ISSUE #387: Exposing config for fields to admin and element SDK
None.

Expose fields to configure fields on admin and showing them on front-end. We should have a PUT / GET.

None.

**Additional context**
None.

segoddnja:
create endpoints to fetch and update
create migration with example config
create test suite
implement basic config validation

closes #387

### DIFF
--- a/cg/migrations/20180815171916_data-attributes-configuration.js
+++ b/cg/migrations/20180815171916_data-attributes-configuration.js
@@ -1,0 +1,35 @@
+const KEY = 'SUBJECT_DATA_ATTRIBUTES_CONFIG';
+
+exports.up = async function(knex, Promise) {
+  await knex('config').insert({
+    key: KEY,
+    value: JSON.stringify({
+      email: {
+        type: 'email',
+        label: 'Your email address',
+        required: true
+      },
+      firstname: {
+        type: 'string',
+        label: 'First name',
+        required: true
+      },
+      lastname: {
+        type: 'string',
+        label: 'Last name',
+        required: true
+      },
+      website: {
+        type: 'string',
+        label: 'Website',
+        required: true
+      }
+    })
+  });
+};
+
+exports.down = async function(knex, Promise) {
+  await knex('config')
+    .where('key', KEY)
+    .delete();
+};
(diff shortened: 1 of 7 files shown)


cg/src/domains/management/data/data.service.js
```diff
@@ -0,0 +1,14 @@
+const KEY = 'SUBJECT_DATA_ATTRIBUTES_CONFIG';
```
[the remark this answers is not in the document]
segoddnja: I disagree. This const is for internal usage only, what is the sense to export it? Only to use it in seeds/migration and tests? Exposing something outside only for testing purpose is a bad practice.

cg/migrations/20180815171916_data-attributes-configuration.js
```diff
@@ -0,0 +1,35 @@
+const KEY = 'SUBJECT_DATA_ATTRIBUTES_CONFIG';
```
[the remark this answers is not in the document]
segoddnja: See my comment below

cg/src/domains/management/data/data.service.js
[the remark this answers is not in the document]
fjcero: Ok, if we are using it only here makes sense to keep it that way.

cg/seeds/example_data_attributes_config.js
```diff
@@ -0,0 +1,27 @@
+exports.seed = async function(knex, Promise) {
```
sindelio: What will this seed be used for?

cg/src/domains/management/management.routes.js
```diff
@@ -146,6 +151,17 @@ module.exports = app => {
     asyncHandler(async (req, res) => usersController.updatePassword(req, res))
   );
 
+  router.put(
```
sindelio: We're mixing REST here, it makes no sense to start doing it now. Either we change everything or keep it as it was before. Please change these endpoints

cg/seeds/example_data_attributes_config.js
fjcero: For creating a default configuration in the Controller side, will be used mainly for demo and elements sdk

sindelio: Oks

cg/src/domains/management/management.routes.js
segoddnja: Well, to some extent this makes sense, although I do not like and idea to do something wrong just to be consistent with something.

PR #404, opened 2018-08-16
Subjects data status on new processor created

### LINKED ISSUE #337: Don't redirect if request fails
**Describe the bug**
The new processors does not have an address, so the request fails. Currently the request fails, but user is still redirected. The subject does not exist. @fjcero Please complete or modify.

**To Reproduce**
1. login to admin.
2. add a new processor.
3. login to front-end demo.
4. consent to all processors.

**Expected behavior**
We should handle error gracefully and show some error message.

**Screenshots**
None.

**Additional context**
None.

fjcero:
Fixes #337 

## Proposed Changes
- Removed console.log
- Some styling to the current subject's data status
- Fixed issues on displaying new processors in subject's data status

### DIFF
--- a/frontend/src/Elements/components/Portability/UserData.js
+++ b/frontend/src/Elements/components/Portability/UserData.js
@@ -23,8 +23,8 @@ class UserData extends React.PureComponent {
 
     if (processors && status.processors) {
       processorsList = processors.map(p => {
-        const s = _.find(status.processors, { id: p.id });
-        p.status = SUBJECT_DATA_STATUS[s.status];
+        const s = _.find(status.processors, { id: p.id }) || {};
+        p.status = s.status ? SUBJECT_DATA_STATUS[s.status] : 'Unconsented';
         return p;
       });
     }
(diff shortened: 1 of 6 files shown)


frontend/src/Elements/components/Portability/UserData.js
```diff
@@ -23,8 +23,8 @@ class UserData extends React.PureComponent {
 
     if (processors && status.processors) {
       processorsList = processors.map(p => {
-        const s = _.find(status.processors, { id: p.id });
-        p.status = SUBJECT_DATA_STATUS[s.status];
+        const s = _.find(status.processors, { id: p.id }) || {};
+        p.status = s.status ? SUBJECT_DATA_STATUS[s.status] : 'Unconsented';
```
segoddnja: I think it makes sense to move processors decoration to Subject context and do it only once - when fetch processors.

fjcero: I like the idea, will do it!

frontend/src/Elements/theme/SubjectData.scss
```diff
@@ -2,6 +2,10 @@
 
 .data {
   .processor {
+    border: 1px solid rgba(0,0,0,0.1);
```
Nelrohd: gonna be picky but should we use variable for colours?

PR #407, opened 2018-08-20
Feature/admin data attributes configurator

segoddnja:
Implement basic admin UI to manage data attributes configuration

Screenshot

### DIFF
--- a/admin/src/containers/Data/AttributesConfigContext.js
+++ b/admin/src/containers/Data/AttributesConfigContext.js
@@ -1,0 +1,67 @@
+import React, { Component, createContext } from 'react';
+import internalFetch from 'helpers/internal-fetch';
+import PropTypes from 'prop-types';
+import AppConfig from 'config';
+
+const AttributesConfigContext = createContext({
+  fetchConfig: () => {},
+  updateConfig: () => {},
+  config: null,
+  isBusy: false
+});
+
+export class AttributesConfigProvider extends Component {
+  static propTypes = {
+    children: PropTypes.oneOfType([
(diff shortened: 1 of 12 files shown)


admin/src/containers/Data/AttributesConfigContext.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  updateConfig = async config => {
+    this.setBusy(true);
+    await internalFetch(`${AppConfig.API_URL}/api/management/data/attributes-config/update`, {
+      method: 'POST',
+      body: JSON.stringify(config)
+    });
+    this.setState({ config });
+    this.setBusy(false);
```
Nelrohd: why not just change the state directly in the line above? this.setState({ isBusy: false }) is clear enough I think.

cg/src/domains/management/data/data.controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+  async getAttributesConfig(req, res) {
+    let config = await this.service.getAttributesConfig();
+    res.json(config);
+  }
+
+  async updateAttributesConfig(req, res) {
+    await this.service.updateAttributesConfig(req.body);
+    res.status(204);
+    return res.send();
```
Nelrohd: no need return

admin/src/containers/Data/AttributesConfigContext.js
segoddnja: That's true. I added this helper only to not type `isBusy` many times in different places, because it is possible to make a typo.

PR #408, opened 2018-08-20
Return the actual state instead of the corresponding number in the request access endpoint

### LINKED ISSUE #397: Return the actual state instead of the corresponding number in the request access endpoint
Return the actual state instead of the corresponding number in the request access endpoints of the subjects

0 - unconsented
1 - consented
2 - erased

None.

**Additional context**
None

segoddnja:
use meaningful string constants for data statuses
closes #397


segoddnja:
@Nelrohd Sure I did. Just one note. I did not change smart contract code, so, in the blockchain statuses are still int numbers, not strings. Mapping happens in the subjects service. Right now make changes in smart contract is a bit difficult, because there is no test suite for it and it's necessary to use some external tools to compile it and put compilation results in the repo. We definitely need to add smart contract compilation to build procedure and do not keep bytecode and ABI in repo.

Issue #398, opened 2018-08-15
Fix the manager's JWT expiring too fast


sindelio:
**Describe the bug**
The management JWT is expiring too fast

**To Reproduce**
1. Start everything with docker/run
2. Log in as a manager and get a JWT
3. Wait 5 minutes
4. JWT will have expired

**Expected behavior**
The management JWT should last around a day, as the subjects JWT

**Screenshots**
None

**Additional context**
See the JWT code

PR #409, opened 2018-08-20
Add env var and default for jwt expiration

Nelrohd:
Fixes #
#398 

## Proposed Changes
  - Added default 7 days
  - Added possibility to define env var


api/src/utils/jwt.js
```diff
@@ -27,7 +27,9 @@ class JWT {
   }
 }
 
-exports.JWT = JWT;
+const DEFAULT_JWT_TOKEN_EXPIRY = 3600;
```
segoddnja: I see in pr description you mentioned 7 days, but here default expiration time is one hour.

PR #414, opened 2018-08-22
Remove system access from erased subject

### LINKED ISSUE #342: Erased subject has still access to the system
**Describe the bug**
After a subject's data is erased, he still has full access to the system

**To Reproduce**
1. Start up the system
2. Go to Postman
3. in the DEMO API, register a subject. Copy the cgtoken returned to the environment as the subject JWT. (Top right of the screen)
4. Go to the CG API, in the subjects section, give consent for that subject
5. Verify that the subject has given consent with the GET endpoints
6. Erase the subject fro the system
7. Try to read the erased subject's data, get objections, get restrictions and so on

**Expected behavior**
After a subject's data is erased, he should not be able to do anything in the system.

**Screenshots**
None.

**Additional context**
@Nelrohd @segoddnja you guys can think about what an erased subject should be able to do in the system. At first I would say not much.

Nelrohd:
Fixes #342 

## Proposed Changes
  - Added middleware `isSubjectErased` on `subject.routes`.

### DIFF
--- a/cg/src/domains/subjects/subjects.middlewares.js
+++ b/cg/src/domains/subjects/subjects.middlewares.js
@@ -6,7 +6,7 @@ const {
   NotFound
 } = require('../../utils/errors');
 
-const isSubjectErased = async (req, res, next) => {
+const subjectNotErased = async (req, res, next) => {
   const controllerStatus = await getSubjectDataState(req.subject.id);
 
   if (controllerStatus === 2) {
@@ -17,5 +17,5 @@ const isSubjectErased = async (req, res, next) => {
 };
 
 module.exports = {
-  isSubjectErased
+  subjectNotErased
 };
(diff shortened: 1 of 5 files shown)


sindelio:
Maybe we should rename the middleware to 'subjectNotErased', so it would be the same name used in the smart-contract and also an affirmation, not an open question

cg/src/domains/subjects/subjects.middlewares.js
```diff
@@ -0,0 +1,21 @@
...
+} = require('../../utils/blockchain');
+
+const {
+  NotFound
+} = require('../../utils/errors');
+
+const isSubjectErased = async (req, res, next) => {
+  const controllerStatus = await getSubjectDataState(req.subject.id);
+
+  if (controllerStatus === 2) {
```
fjcero: There is a way to replace that `2` by a constant?

PR #419, opened 2018-08-24
add postman collections and environment to repository

segoddnja:


cg/docs/postman/CG.postman_environment.json
```diff
@@ -0,0 +1,42 @@
...
+  "values": [
+    {
+      "key": "subject_jwt",
+      "value":
+        "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWJqZWN0SWQiOjEsImlhdCI6MTUzMjU2NDc0Mn0.lXhu6igKzCo--Vb6jjFoknYdoinBzMZqv_ze6hfqWpE",
+      "enabled": true,
+      "type": "text"
+    },
+    {
+      "key": "og_base_url",
```
sindelio: Maybe we should clear some fields in the environment?

sindelio:
I checked why the build is breaking, @Nelrohd the test you made for the race condition is breaking it. You have to unmock the function after it is used. Maybe use mockOnce? 
Also, please check the state of the database for the test to be complete: 
```const subjectExists = await this.db('subjects').where('id', subjectId);
expect(subjectExists).toBe(null or undefined);```

Nelrohd:
@sindelio no the test was using the previous URL of giving consent. See https://github.com/ClearGDPR/ClearGDPR/pull/423

PR #424, opened 2018-08-27
add management route, controller action and service method to allow m…

segoddnja:


cg/src/domains/management/subjects/subjects.service.js
```diff
@@ -65,7 +68,22 @@ class SubjectsService {
...
+      .join('subject_keys', 'subjects.id', '=', 'subject_keys.subject_id')
+      .select('personal_data')
+      .select('key')
+      .where({ subject_id: subjectId });
+
+    if (!data) {
+      throw new NotFound('Subject not found');
+    }
+    const decryptedData = decryptFromStorage(data.personal_data, data.key);
+    await recordAccessByController(subjectId);
```
sindelio: This records the access from a subject to it's own data, not the manager's access

segoddnja: Why the method name is `recordAccessByController` in this case?

sindelio: To differentiate between controller and processor actions

sindelio:
Since it's done already, we should make use of it. In the future if the client don't want this feature, we can remove it. Please put it in a passing state @segoddnja and then we can merge

Issue #370, opened 2018-08-02
Publish Elements SDK as standalone package


fjcero:
**Is your feature request related to a problem? Please describe.**
- [x] Move Elements SDK outside the frontend package
- [x] Add boilerplate to Elements SDK (prettier, jest, eslint, build scripts)
- [ ] Create the build scripts
- [ ] Add an example on how to instantiate Elements SDK using the package or CDN

For next sprint
- [ ] Make a publish script to NPM
- [ ] Automatically deploy when a new version is released (tagged on Github)

**Describe the solution you'd like**
The main objective is to allow developers to easily interact with Elements component, so the import should be done with the least amount of code from the dev side.

**Describe alternatives you've considered**
Instead of moving outside of frontend, perhaps lerna could be used, also, makes sense to start using tools to manage monorepo packages.

**Additional context**
Should be addressed after major chances are introduce

PR #427, opened 2018-08-28
#370 Moved JS-SDK to packages/

fjcero:
Start using Lerna management to share `js-sdk` across packages. 

- Added config files for the package
- Added build script configuration
- Split Events handling into a new file


fjcero:
@segoddnja you're right, just put it for testing purposes, should be removed.

